Issue #4, opened 2016-10-25
This module would also support the HS200 wall switch if you would include it in models.


smaggard:
I modified the latest code locally on my home assistant server and it worked perfectly with a hs200 wall switch.  It works exactly as an HS100 switch

def _identify_model(self):
        """Query sysinfo and determine model"""
        sys_info = self.get_info()

```
    if sys_info["system"]["get_sysinfo"]["model"][:5] == 'HS100':
        model = 100
    elif sys_info["system"]["get_sysinfo"]["model"][:5] == 'HS200':
        model = 100
    elif sys_info["system"]["get_sysinfo"]["model"][:5] == 'HS110':
        model = 110

    return model
```

smaggard:
I have a branch created with full addition of the HS200 switch if you could a add me as a contributor.

smaggard:
Code changes I am proposing can be found in https://github.com/smaggard/pyHS100/blob/master/pyHS100.py

GadgetReactor:
That's great.

Please start a new branch, raise a pull request and we will be able to
check and merge it in.

smaggard:
I have a new branch but I can't publish it

smaggard:
I am getting the following error trying to create a branch, I think I need some permissions set.

You may not have permission to access pyHS100. Check Preferences to make sure you’re still logged in.

smaggard:
I tried doing a pull request and got a permission denied that route too, let me know when you get the permissions setup.

GadgetReactor:
Hi, there isn't any permissions. You won't be able to publish directly. Please see how to below, it's the same guidelines as used by Home-Assistant. 

## How to contribute

### If doing this for the first time

Fork this repo using the Fork button.
Clone your fork on your computer.

git clone https://github.com/YOUR_GITHUB_USERNAME/pyHS100.git
Add the main repo as upstream.

cd pyHS100
git remote add upstream https://github.com/gadgetreactor/pyHS100.git

### Every time

Be sure to always work with the current upstream master.

### Run this from your feature branch

git fetch upstream master  # to pull the latest changes into a local branch
git rebase upstream/master # to put those changes into your feature branch before your changes
Create a feature branch for your changes. This is important - don't push to master!

git checkout -b new-feature
Also, do a special branch based on your master for each feature you want to implement - don't put multiple features into one branch - this is also very important!

[...]

PR #5, opened 2016-10-26
Adding in support for the HS200 Wall switch referencing issues (#4)

smaggard:


GadgetReactor (approved):
Thanks for making the changes.

PR #8, opened 2016-11-21
Externalize the TP-Link Smart Home Protocol

mweinelt:
Signed-off-by: Martin Weinelt <hexa@darmstadt.ccc.de>

1) The Smart Home Protocol does not require access to members of the SmartPlug class and is an underlying part realizing the network communication, while the SmartPlug represents a Plug itself.

2) The device advertises the features it offers in `get_sysinfo`, so use that instead of vaguely matching against model detection. Also the implementation was broken, because `re.sub()` would return a `str` which was being compared against an `int`.

3) Normalizes all docstrings, add SWITCH_STATE_ON/OFF/UNKNOWN constants, representing switch states. Express commands with native python types.


GadgetReactor:
merged

PR #16, opened 2016-12-12
Version bump

kirichkov:


kirichkov:
I've set the version to 0.2.2 but perhaps 0.3 is more fitting?

rytilahti:
Btw, would it be time to do a proper tagged release this time?

Anyway, I'd like to get some code cleanups (splitting the module to more files, it's pretty huge for one file already) done + maybe prepare the codebase for mockable backend for testcases. Do you think it could go into the next release, or shall it wait for the next one?

GadgetReactor:
The next version can be 0.3

kirichkov:
@rytilahti My personal opinion is that before adding new features it's better to refactor the code and write the new code in the "proper" way and with tests added in. I think the most important part here was to have a new version, that can go into home assistant to fix the issue with the HS200 switches.

@mweinelt Have the bugs that required a home assistant revert been fixed? Can we get this version into home-assistant to restore the HS200 functionality?

rytilahti:
@kirichkov Ok, I understand. I did some refactoring and added tests (#17), based heavily on HS110 info as that's all I have. So the code now partially testable (and that it already revealed a bug in error-handling and problems with python2 compatibility, which are also now fixed).

mweinelt:
@kirichkov Yes, they have been, but the API changed a bit, so the bindings need updating now.

kirichkov:
OK. I'll work on it.

PR #20, opened 2016-12-20
allow manual updates of the smartplug state

mweinelt:
Signed-off-by: Martin Weinelt <hexa@darmstadt.ccc.de>

The home-assistant bindings want to update the smartplug state in intervals, since there are other applications that could interact with the smartplug.

Let's make the method public and call it aptly `update`.

Also bump version to 0.2.3, since this warrants a quick update to roll out to all homeassistant users.

references:
https://github.com/home-assistant/home-assistant/issues/4992
https://github.com/home-assistant/home-assistant/pull/4914

### DIFF
--- a/pyHS100/pyHS100.py
+++ b/pyHS100/pyHS100.py
@@ -75,16 +75,6 @@ class SmartPlug:
         self.protocol = protocol
         self._sys_info = None
 
-    def _fetch_sysinfo(self):
-        """
-        Fetches the system information from the device.
-
-        This should be called when the state of the plug is changed.
-
-        :raises: SmartPlugException: on error
-        """
-        self._sys_info = self.get_sysinfo()
-
     def _query_helper(self, target, cmd, arg={}):
         """
         Helper returning unwrapped result object and doing error handling.
@@ -114,10 +104,20 @@ class SmartPlug:
 
         return result
 
+    def update(self):
+        """
+        Fetches the system information from the device.
+
+        This should be called when the state of the plug is changed.
+
+        :raises: SmartPlugException: on error
+        """
+        self._sys_info = self.get_sysinfo()
+
     @property
     def sys_info(self):
         if not self._sys_info:
-            self._fetch_sysinfo()
+            self.update()
 
         return self._sys_info
 
@@ -163,7 +163,7 @@ class SmartPlug:
         else:
             raise ValueError("State %s is not valid.", value)
 
-        self._fetch_sysinfo()
+        self.update()
 
     def get_sysinfo(self):
         """
@@ -202,7 +202,7 @@ class SmartPlug:
         """
         self._query_helper("system", "set_relay_state", {"state": 1})
 
-        self._fetch_sysinfo()
+        self.update()
 
     def turn_off(self):
         """
@@ -212,7 +212,7 @@ class SmartPlug:
         """
         self._query_helper("system", "set_relay_state", {"state": 0})
 
-        self._fetch_sysinfo()
+        self.update()
 
     @property
     def has_emeter(self):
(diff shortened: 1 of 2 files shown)


rytilahti:
As mentioned in https://github.com/home-assistant/home-assistant/pull/4914 , I'd prefer this to be handled transparently to the apps, by either 1) letting sys_info property fetch sysinfo for each call or 2) have a cache for a second or two if it'd otherwise spam the devices too much.

We need more tests for the interoperability too to avoid this happening again. Unfortunately I'll be away and can't do much about this until later on, so if a hotfix is warranted I'd propose to do it by calling the _fetch_sysinfo().

edit: however, if we proceed with having a cache, we should provide a way to do a full reset of the info, so adding reset() or similarly named method to force state refresh should get included.

kirichkov:
This being a communication library I don't think we need to worry about hammering the devices. This is a responsibility of the code using this library (e.g. home assistant) and letting it do as its developer desires.

I think the lowest level class should be as simple as possible and not try to cache anything that can change in time, i.e. it's OK to cache hardware and firmware versions but not state and schedule.

rytilahti:
@kirichkov so, what's your opinion, would it make then sense to just adapt sysinfo getter to call fetch sysinfo under the hood? We can adapt those special getteers for hw/fw do caching if needed afterwards.

kirichkov:
I'm actually fine with @mweinelt's solution. And now that I think of it, let's assume the following imaginary scenario:

1. Plug API changes between version 2.0 and 3.0
2. User updates their plug
3. User must restart home assistant (or any other software using this module) so that the SW version can be re-fetched and the proper API used.

I'd say let's not cache anything. Let the consumers (e.g. Home assistant) decide what they want cached, and what not.

@GadgetReactor can you get this merged?

rytilahti:
I completely agree that we should not cache anything, and therefore I disagree with merging this to be honest, it just hides the problem without fixing it :)

So what I'm trying to propose is the following:
1) Call _fetch_sysinfo() in sysinfo and return its value
2) The parsing logic (e.g. splitting features) gets moved out from wrappers towards _fetch_sysinfo(), this way users accessing sysinfo object directly still get the benefits of parsed information even if accessing the "raw" sysinfo object.
3) Document the behavior, that those getter wrappers fetch the state from the device, in case this is _really_ problematic, one just uses sysinfo()'s and read elements from it directly.

This way we can avoid introducing update(), we still get the benefits of parsed "sysinfo" for those getters, and there are clean boundaries where the device is being accessed (that is, by every getter, by every setter, no need for magical "update" to keep the cache up to date.)

rytilahti:
I created a proof-of-concept of what I'm proposing: https://github.com/rytilahti/pyHS100/commit/4328ff895fd58109e0911d4ea17e06f1766f35b8

kirichkov:
Some questions:

https://github.com/rytilahti/pyHS100/blob/remove-caching/pyHS100/pyHS100.py#L192
A lame question because I have mediocre python knowledge: What would happen here if `self.sys_info['relay_state']` returns `None`, instead of a boolean? Also I remember there being differences in Python 2 and Python 3 with None's "truthyness".

https://github.com/rytilahti/pyHS100/blob/remove-caching/pyHS100/pyHS100.py#L293
Isn't it better to rescue the exception and return False?

rytilahti:
To answer to the first question, there's a bool() cast in-place in is_on(), so None would be turned to False. not False will be True then, seems to work at least with py3.5.2 and py2.7.13 as expected. However, using voluptuous/other way of parsing & enforcing the format, or at least informing user about inconsistencies coming from the devices would, so in the end it shouldn't ever be "None".

Catching an exception and returning false: it'll hide or make it difficult to see where the error is coming from. And it's more consistent, when error reporting follows the same scheme by raising exceptions.

GadgetReactor:
I would think that it is better to leave the caching mechanism (if required) to be at Home-Assistant end. But my knowledge is really limited, so I would defer to you guys.

rytilahti:
Created #26, if we decide this is the way to go, it'd be nice if someone else could also test the code.

mweinelt:
Will have to be rebased onto the SmartBulb changes if we want to go this road.

PR #23, opened 2016-12-31
Ignore error when closing read side of socket after closing write side

harmingcola:
On Mac OS X, shutdown(SHUT_RD) fails if the write side on the other side has been closed.

### DIFF
--- a/pyHS100/protocol.py
+++ b/pyHS100/protocol.py
@@ -53,9 +53,10 @@ class TPLinkSmartHomeProtocol:
 
         try:
             sock.shutdown(socket.SHUT_RDWR)
-            sock.close()
-        except:
+        except OSError:
             pass
+        finally:
+            sock.close()
 
         response = TPLinkSmartHomeProtocol.decrypt(buffer[4:])
         _LOGGER.debug("< (%i) %s", len(response), response)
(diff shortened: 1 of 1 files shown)


rytilahti (changes requested):
Please avoid using "catch all" except: https://docs.python.org/2.7/howto/doanddont.html#except

Also please add a short comment why exceptions are pass'd.

dims:
I've tested this change. works for me!

mweinelt:
Should probably move `sock.close()` into finally and <s>maybe downgrade to `sock.shutdown(socket.SHUT_RD)`</s>.

```python
try:
    sock.shutdown(socket.SHUT_RDWR)
except OSError:
    pass
finally:
    sock.close()
```

rytilahti:
@GadgetReactor any news on this?

rytilahti (approved):

mweinelt:
@harmingcola Can you please rebase this PR against master? We'll get it merged quickly then.

mweinelt:
I'm not sure we still need this after the latest protocol changes after all.

rytilahti:
Do you have some specific reason for that? From what I see the original problem is caused by calling shutdown() on a closed socket on OSX which are not at least directly affected by those protocol changes.

mweinelt:
Obsoleted in #41

Issue #14, opened 2016-12-04
Issue receiving state of switch


hoveeman:
I updated home assistant last night to .34 and it pulled down the latest pyHS100. My hs200 wall switches are now able to be controlled which is great! However, turning a hs100/hs110/hs200 on from home assistant turns the light on, but after a second, home assistant toggle flips to off. The light switch physically still stays on though. Reloading the ha interface shows the switch as off. I can tap to toggle it on and the switch physically stays on. If I tap it again to turn it off, it will turn off the switch. It appears as if it is unable to poll the device to receive its state after the last update.

nickhx:
exact same problem here with my HS100 (UK) - sending a turn off or turn on command via home assistant works.  However, the toggle command turns on from the off state, but does not turn off again when turned on.

smaggard:
I think is caused by the change in the code here
`        if relay_state == 0:
            return SWITCH_STATE_OFF
        elif relay_state == 1:
            return SWITCH_STATE_ON
        else:
            _LOGGER.warning("Unknown state %s returned.", relay_state)
            return SWITCH_STATE_UNKNOWN`

That method used to return simply 'ON', 'OFF', or 'UNKNOWN'  which home assistant could handle.

nickhx:
@smaggard well done - the problem is that the SWITCH_STATE_OFF and SWITCH_STATE_ON variables are now defined as lowercase 'on' and 'off'.  In Home Assistant,  tplink.py is expecting uppercase 'ON' or 'OFF'.  So, should this library be updated or is it better to update tplink.py in Home Assistant?

smaggard:
I think it would be easiest for the contributors currently making changes to simply change the values of the variables here, but I will leave that up to them.

smaggard:
The following would fix the issue.

`        if relay_state == 0:
            return SWITCH_STATE_OFF.upper()
        elif relay_state == 1:
            return SWITCH_STATE_ON.upper()
        else:
            _LOGGER.warning("Unknown state %s returned.", relay_state)
            return SWITCH_STATE_UNKNOWN.upper()`

mweinelt:
You are looking at the wrong versions. This is clearly before the changes you're mentioning.

Home-Assistant uses 0.2.1, which is based on https://github.com/GadgetReactor/pyHS100/commit/53c06ca7844f16d6ffe068f4ca01e8d5c707b9e3.

[...]

PR #24, opened 2017-01-05
Make tests runnable without device

rytilahti:
Adds preliminary support for fake devices, thanks to
hoveeman's sysinfos from issue #14,
making running tests possible without a device.

At the moment we have only HS110 and HS200 infos available, and tests
are currently run only against HS110 data.

### DIFF
--- a/pyHS100/pyHS100.py
+++ b/pyHS100/pyHS100.py
@@ -14,8 +14,13 @@ You may obtain a copy of the license at
 http://www.apache.org/licenses/LICENSE-2.0
 """
 
+# python2 compatibility
 from __future__ import absolute_import
 from __future__ import unicode_literals
+try:
+    basestring
+except NameError:
+    basestring = str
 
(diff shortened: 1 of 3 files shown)


rytilahti:
Oh well, supporting even py27 requires some effort it seems. Now the tests do run with py27, py34 and py35 :-) The question I'm thinking is whether it's worth to try to keep py27 compatibility at all, opinions?

kirichkov:
Given this is primarily used for Home Assistant, I think python3 should be priority. But python2 is a lot more used. I'd say don't concentrate on it, as the code is getting more and more complex and I'd rather have new features added to Home Assistant than waste time on keeping it 2.7 compliant.

rytilahti:
Yeah, I'll let it stay there for now as it's not that much work with automated tests. If it's considered too much, it should be pretty easy to get rid of that cruft later on.

Issue #9, opened 2016-11-22
Device Discovery via TDDP


mweinelt:
https://www.softscheck.com/en/reverse-engineering-tp-link-hs110/#TP-Link Device Debug Protocol

It says the TP-Link Device Debug Protocol has the ability to discover devices inside the same broadcast domain. A reference implementation of the general protocol is available at https://github.com/softScheck/tplink-smartplug/tree/master/tddp-client.

GadgetReactor:
Ah thanks so much for this!

mweinelt:
Also check out this fresh vulnerability report: http://www.coresecurity.com/advisories/tp-link-tddp-multiple-vulnerabilities

rytilahti:
Device discovery doesn't need TDDP implementation, just a broadcast UDP packet to 255.255.255.255:9999. The devices will reply with usual payloads.

Request payload:
{"system": {"get_sysinfo": None}, "emeter": {"get_realtime": None}}

rytilahti:
This has been implemented (not with TDDP though, but I'd say extra work on this is not worth the effort), so closing for now.

babanaweed:
Please help me with this exploit how to use it.
https://www.exploit-db.com/exploits/40814/

PR #25, opened 2017-01-05
Add device discovery

rytilahti:
These commits add ability to discover devices in the local network, and adds an example script showing its usage. Doesn't use TDDP, but fulfils #9 for local discovery.

The code has been tested with HS110.


kirichkov:
👍

kirichkov:
I'm not able to get this working under raspbian and Ubuntu. What OS did you use to get successful discovery?

I'm clearly seeing the response from the plug in tcpdump, but the discovery times out.

rytilahti:
Raspbian here. The discovery is supposed to time out (it's udp, so we have to quit listening for responses at some point), but it should print out the results after that.

kirichkov:
I'm getting this, but I clearly see the response from the Plug in tcpdump. It's important to note that my plug is on a separate VLAN and my hass box also has a vlan interface on the same VLAN, this might be related.

```
>>> from pyHS100 import TPLinkSmartHomeProtocol
>>> TPLinkSmartHomeProtocol.discover()
Got exception timed out
  File "/home/kirichkov/dev/hass/pyHS100/pyHS100/protocol.py", line 99, in discover
    data, addr = sock.recvfrom(4096)
socket.timeout: timed out
[]
>>> 
```

rytilahti:
It works only inside one network, broadcasts are not forwarded to other networks by routers.

kirichkov:
While investigating non-related issue, I think a socket.bind("ip_of_vlan") would fix my issue. I think why it's failing is because python attaches to the first interface listening for responses, and not to all interfaces.

rytilahti:
You can't receive broadcasts between vlan segments/networks, that's just how it works with routers. If it is in the same network (even from many), you can try if sock.bind("", <port>) helps. That'll bind on all available interfaces.

kirichkov:
Nope, what you suggest does not work. If I do bind on the vlan interface's IP address, it works!

PR #26, opened 2017-01-08
Initial steps to remove caching

rytilahti:
This commit removes caching of sysinfo to avoid
inconsistent states as described in issue #14.

Each an every access for properties will cause a request
to be made to the device. To avoid this, user of the library
may want to access sys_info() directly instead of using the helpers.

Currently sys_info() returns raw json object where-as helpers do
parse information for easier consumption; current state is just to
provide a PoC how it looks compared to having an active update()
for fetching the info.

### DIFF
--- a/pyHS100/pyHS100.py
+++ b/pyHS100/pyHS100.py
@@ -80,17 +80,6 @@ class SmartPlug(object):
         if not protocol:
             protocol = TPLinkSmartHomeProtocol()
         self.protocol = protocol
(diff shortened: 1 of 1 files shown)


kirichkov:
I've tested this. The re-quering works and UI gets updated.

@GadgetReactor I think it's ready for merging. 
Please include a version bump in `setup.py` so I can update the dependency in HASS.

PR #30, opened 2017-01-14
Add support for TP-Link smartbulbs

mjg59:
This patchset adds support for TP-Link smartbulbs running the latest firmware. There appears to have been a slight modification to the protocol - the previously empty header is now the length of the request. It's possible that the emeter changes are firmware-version dependent rather than specific to the bulbs, so that may need to be refactored when we get more feedback.

### DIFF
--- a/pyHS100/__init__.py
+++ b/pyHS100/__init__.py
@@ -1,3 +1,3 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
-from pyHS100.pyHS100 import SmartPlug, TPLinkSmartHomeProtocol, SmartPlugException
+from pyHS100.pyHS100 import SmartBulb, SmartPlug, TPLinkSmartHomeProtocol, SmartPlugException
--- a/pyHS100/pyHS100.py
+++ b/pyHS100/pyHS100.py
@@ -38,39 +38,10 @@ class SmartPlugException(Exception):
     pass
 
 
-class SmartPlug(object):
-    """Representation of a TP-Link Smart Switch.
-
-    Usage example when used as library:
-    p = SmartPlug("192.168.1.105")
-    # print the devices alias
-    print(p.alias)
-    # change state of plug
-    p.state = "ON"
-    p.state = "OFF"
-    # query and print current state of plug
-    print(p.state)
-
-    Errors reported by the device are raised as SmartPlugExceptions,
-    and should be handled by the user of the library.
-
-    Note:
-    The library references the same structure as defined for the D-Link Switch
-    """
-    # switch states
-    SWITCH_STATE_ON = 'ON'
-    SWITCH_STATE_OFF = 'OFF'
-    SWITCH_STATE_UNKNOWN = 'UNKNOWN'
-
-    # possible device features
-    FEATURE_ENERGY_METER = 'ENE'
-    FEATURE_TIMER = 'TIM'
-
-    ALL_FEATURES = (FEATURE_ENERGY_METER, FEATURE_TIMER)
-
+class SmartDevice(object):
     def __init__(self, ip_address, protocol=None):
         """
-        Create a new SmartPlug instance, identified through its IP address.
+        Create a new SmartDevice instance, identified through its IP address.
 
         :param str ip_address: ip address on which the device listens
         :raises SmartPlugException: when unable to communicate with the device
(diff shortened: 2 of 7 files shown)


pyHS100/__init__.py
```diff
@@ -1,3 +1,3 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
-from pyHS100.pyHS100 import SmartPlug, TPLinkSmartHomeProtocol, SmartPlugException
+from pyHS100.pyHS100 import SmartBulb, SmartPlug, TPLinkSmartHomeProtocol, SmartPlugException
```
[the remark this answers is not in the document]
mweinelt: Ignore this.

pyHS100/pyHS100.py
```diff
@@ -37,37 +37,7 @@ class SmartPlugException(Exception):
...
-    SWITCH_STATE_OFF = 'OFF'
-    SWITCH_STATE_UNKNOWN = 'UNKNOWN'
-
-    # possible device features
-    FEATURE_ENERGY_METER = 'ENE'
-    FEATURE_TIMER = 'TIM'
-
-    ALL_FEATURES = (FEATURE_ENERGY_METER, FEATURE_TIMER)
-
+class SmartDevice(object):
```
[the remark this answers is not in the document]
mjg59: I can't figure out what this complaint is - there appear to be two blank lines after this patch.

mjg59:
The Python 2 CI failure doesn't seem to be related to any of my changes?

mweinelt:
No, they're related to the "raise from" Syntax, which is not supported in Python 2. We'll have to use the implementation in six or future instead.

mjg59:
Ok. Any other changes you'd like? I think the remaining hound issue is spurious.

pyHS100/tests/test_protocol.py
```diff
@@ -10,6 +10,6 @@ class TestTPLinkSmartHomeProtocol(TestCase):
     def test_encrypt(self):
         d = json.dumps({'foo': 1, 'bar': 2})
         encrypted = TPLinkSmartHomeProtocol.encrypt(d)
-        # encrypt appends nullbytes for the protocol sends
-        encrypted = encrypted.lstrip(b'\0')
+        # encypt adds a 4 byte header
```
mweinelt: typo, encrypt

mweinelt:
I'd like some feedback from @rytilahti and @kirichkov.

pyHS100/pyHS100.py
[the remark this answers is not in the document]
mweinelt: Looks good to me as well, the SmartDevice class could use a docstring though.

pyHS100/__init__.py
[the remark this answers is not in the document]
rytilahti: #31 will fix this.

pyHS100/pyHS100.py
```diff
@@ -534,3 +279,477 @@ def mac(self, mac):
...
+            bulb_state = 0
+
+        light_state = {
+            "on_off": bulb_state,
+        }
+        return self.set_light_state(light_state)
+
+    @property
+    def has_emeter(self):
+        return True
```
mweinelt, line 617: Does the bulb expose a feature list in its sysinfo? If so, please check for emeter in that list.

rytilahti:
Unit-tests run fine here, examples.cli works with HS110. It'd probably make sense to split up pyHS100.py to SmartBulb, SmartPlug and SmartDevice, the file is getting big.

@mjg59 may we get also unit-tests for bulb functionality?
edit: please also add contents of sysinfo to unit-test file.

@trainman419 do you mind testing if these changes make HS105 work too?

pyHS100/pyHS100.py
```diff
@@ -534,3 +279,477 @@ def mac(self, mac):
...
+    BULB_STATE_OFF = 'OFF'
+
+    def __init__(self, ip_address, protocol=None):
+        SmartDevice.__init__(self, ip_address, protocol)
+        self.is_color = self.sys_info['is_color']
+        self.is_dimmable = self.sys_info['is_dimmable']
+        self.is_variable_color_temp = self.sys_info['is_variable_color_temp']
+        self.emeter_type = "smartlife.iot.common.emeter"
+        self.emeter_units = True
+
```
rytilahti, line 431: Define is_color, is_dimmable and is_variable_color_temp as properties, and skip accessing sysinfo here. Allows lazy initialization in, e.g., home assistant.

pyHS100/pyHS100.py
mjg59: It doesn't, annoyingly. As far as I can tell all the bulbs have this feature.

pyHS100/pyHS100.py
mjg59: That was my initial approach, but it seemed like it'd lead to more network activity in general.

rytilahti: That's true, #26 changed the behaviour to disable caching of responses from the bulb so that every access to sys_info will trigger a request. I'm having second thoughts if it'd make more sense to revert that and make clients to adapt to #20.

pyHS100/pyHS100.py
```diff
@@ -38,36 +38,7 @@ class SmartPlugException(Exception):
...
-    # possible device features
-    FEATURE_ENERGY_METER = 'ENE'
-    FEATURE_TIMER = 'TIM'
-
-    ALL_FEATURES = (FEATURE_ENERGY_METER, FEATURE_TIMER)
-
+class SmartDevice(object):
     def __init__(self, ip_address, protocol=None):
         """
         Create a new SmartPlug instance, identified through its IP address.
```
kirichkov: Change the comment to reflect the change in the class name, i.e. SmartDevice

kirichkov (approved):

kirichkov:
I also think it's high time we split up pyHS100.py into several smaller files.

Also I think this module needs a name change badly to properly reflect the devices it supports, but this should be low on the priority list.

rytilahti:
@mjg59 could you also please implement features in some form, even if it's not reading from "features" but based on some other keys? Those hound checks against fakes.py you can ignore, #31 will add a proper ignore when it gets merged.
edit: those can also be fixed, but I would not consider over-length lines a problem especially with fake test data..

InblEric:
Probably worth posting here as well but this PR did make my HS105 work.

mweinelt:
Huh, this merge was unfortunately a bit premature.

InblEric:
Did my comment trigger an automatic merge somehow? I just wanted to comment about it helping for the HS105 too.

kirichkov:
@InblEric no, don't worry about it! Merges are still manual here :-)

GadgetReactor:
Oops. Should I reverse this?

trainman419:
I've confirmed that this works with my HS105 too!

mweinelt:
@GadgetReactor  Well, the code needs to be updated for PEP8 compliance, see @houndci-bot comments.

Also we didn't get the split into another module, this leaves `pyHS100.py` a 800 LOC monster. Should also be addressed imo.

PR #32, opened 2017-01-14
Try to resolve hostname before opening a socket

craigcabrey:
If `ip_address` is already an IP address, `gethostbyname()` will simply return it unmodified.

### DIFF
--- a/pyHS100/pyHS100.py
+++ b/pyHS100/pyHS100.py
@@ -75,6 +75,7 @@ class SmartPlug(object):
         :param str ip_address: ip address on which the device listens
         :raises SmartPlugException: when unable to communicate with the device
         """
+        ip_address = socket.gethostbyname(ip_address)
         socket.inet_pton(socket.AF_INET, ip_address)
         self.ip_address = ip_address
         if not protocol:


rytilahti (approved):
Looks fine to me.

kirichkov (approved):
👍

mweinelt:
If we start accepting hostnames, lets rename the function parameter to `host`.

Besides that, https://github.com/GadgetReactor/pyHS100/pull/32/files#diff-3475cfc85c1bad6f0bd35e52795d77e8R79 is intended to validate ip addresses, if we pipe them through `gethostbyname`, what purpose does it still serve?

mweinelt (changes requested):
Rename the functions `ip_address` parameter to host

kirichkov:
https://github.com/GadgetReactor/pyHS100/pull/32/files#diff-3475cfc85c1bad6f0bd35e52795d77e8R79 Line 79 should indeed be removed, however, I just saw that `gethostbyname` is not IPv6 compliant. It's recommended to use `getaddrinfo`. So perhaps I was a bit too quick to approve this change.

craigcabrey:
Ah, I didn't realize that the socket wasn't actually for communication, just validation. Good catch :+1:. I'll play around with `getaddrinfo` to see how it behaves in IPv4-only, IPv6-only, and mixed environments.

rytilahti:
I'm wondering whether this is really necessary?

mweinelt:
It's no harm doing dns lookups, and in a more organized automation setup working with hostnames might be desirable. So I'm in favor for this functionality.

rytilahti:
Fair enough indeed, it needs to be ported to use getaddrinfo before getting merged.

mweinelt:
@craigcabrey Are you interested in resolving the outstanding issues?

kirichkov:
I believe this has gone stale.

@craigcabrey are you interested in finishing this up,or should I close the pull request?

kirichkov:
This has gone stale. Closing it. Please reopen when ready.

PR #36, opened 2017-02-11
set color temp to 0 when trying to change color

pete1450:
Attempting to change the hsv while a bulb is currently using a non-zero white color temp results in the values being set but they don't take. Color temp appears to take precedence unless it is 0. By setting color temp to 0 every time hsv is set it should take. I say "should" as I discovered this while translating to perl and haven't tested it in python. Once I got it working in perl I wondered how I missed it and I noticed you didn't have it.


rytilahti:
I don't have a bulb to test, but this looks harmless enough.

PR #49, opened 2017-04-13
Refactor and drop py2 support

rytilahti:
This PR refactors SmartBulb and SmartPlug into their own classes while still keeping the backwards compatibility with the public API (which may change at later point if the module gets renamed).

Besides that python2 support is dropped completely and the code is cleaned where necessary, which should not be a problem as we have never had python2 downstreams and I think there's nothing lost if we do this. This is an effort to make the code-base cleaner for the future as well as allow easier bulb client implementation.


kirichkov (approved):
Looks good!

rytilahti:
Ok, merging it in the hopes that people will try it out before making a release at some point in the future. For the new release we want to make the cli tool to function for bulbs too!

Issue #53, opened 2017-04-24
Set timezone functionality


jeffsf:
In reviewing the code I came across
```
@time.setter
    def time(self, ts):
    raise NotImplementedError("Fails with err_code == 0 with HS110.")
```
and the associated notes that it does not appear to be functioning properly. 

It does not surprise me that the time, itself, can't be set, after watching how aggressively an HS-110 tries to find an NTP server, cycling through around a dozen different host names.

On the other hand, I "guessed" that index 6 was the proper timezone to use at my location and it did accept the change, both in terms of changing the timezone returned, as well as the time offset.

I don't have a list of the zones, which would be, shall we say, "helpful" in putting together a timezone setter. If this is enough of a desire, I can try to determine what the options are, hopefully without bricking my HS-110...

I don't know the relative priority for this, as not all users are adverse to using the TP-Link phone app to configure their devices. At least for me, it's one less thing I have to worry about, especially given the recent revelations about Uber's iPhone app and TP-Link's history of backdoors.

jeffsf:
```
{'index': 0, 'zone_str': '(UTC-12:00) International Date Line West', 'tz_str': '<GMT+12>12', 'dst_offset': 0}
{'index': 1, 'zone_str': '(UTC-11:00) Coordinated Universal Time-11', 'tz_str': '<GMT+11>11', 'dst_offset': 0}
{'index': 2, 'zone_str': '(UTC-10:00) Hawaii', 'tz_str': 'HST10', 'dst_offset': 0}
{'index': 3, 'zone_str': '(UTC-09:00) Alaska', 'tz_str': 'AKST9AKDT,M3.2.0,M11.1.0', 'dst_offset': 60}
{'index': 4, 'zone_str': '(UTC-08:00) Baja California', 'tz_str': 'PST8PDT,M3.2.0,M11.1.0', 'dst_offset': 60}
{'index': 5, 'zone_str': '(UTC-08:00) Pacific Standard Time (US & Canada)', 'tz_str': 'PST8', 'dst_offset': 0}
{'index': 6, 'zone_str': '(UTC-08:00) Pacific Daylight Time (US & Canada)', 'tz_str': 'PST8PDT,M3.2.0,M11.1.0', 'dst_offset': 60}
{'index': 7, 'zone_str': '(UTC-07:00) Arizona', 'tz_str': 'MST7', 'dst_offset': 0}
[...]

PR #55, opened 2017-04-25
Add timezone setter

jeffsf:
While an HS110 may ignore the time in the setter,
it will allow the timezone to be changed.

https://github.com/GadgetReactor/pyHS100/issues/53

### DIFF
--- a/pyHS100/pyHS100.py
+++ b/pyHS100/pyHS100.py
@@ -270,6 +270,26 @@ class SmartDevice(object):
         """
         return self._query_helper("time", "get_timezone")
 
+    @timezone.setter
+    def timezone(self, zone_index):
+        """
+        Set the timezone, taking the current time from the device
+        The HS110, at least, requires date/time, but appears to ignore it
+
+        :param zone_index: int
+        """
+        device_time = self.time
+        tz_set_dict = {
+            'year': device_time.year,
+            'month': device_time.month,
+            'mday': device_time.day,
+            'hour': device_time.hour,
+            'min': device_time.minute,
+            'sec': device_time.second,
+            'index': zone_index
+        }
+        self._query_helper("time", "set_timezone", tz_set_dict)
+
     @property
     def hw_info(self):
         """


kirichkov (changes requested):
Can you please add a file TIMEZONES / TIMEZONES.md that lists all the time zones with their respective index, so that it's easier for newcomers to figure it out instead of digging through the github issues to find the index for their time zone.

A reference from README.md will also be good I think.

jeffsf:
Sounds good -- I'll get the raw output into something a little easier to read. 

I had, at first, thought that it could easily be generated with a simple `for` loop, but the HS105 I literally just opened up isn't as descriptive
```
tp105.timezone
{'index': 6}
```

rytilahti:
I'd much prefer to have this as a part of setter for time, where it imo API-wise belongs.

jeffsf:
@rytilahti -- Since it doesn't seem to set the time, but just the time zone, at least with an HS110, I chose to have it as `@timezone.setter` rather than `@time.setter`

I'm wrestling with an HS105 and will take a look to see if it sets the time, hopefully this evening (Pacific Time). If so, then it makes sense to me to have _both_ a function to set time, as well as one that only sets the time zone.

rytilahti:
@jeffsf yeah, I made that NotImplemented there exactly as it wasn't updating the time. I didn't however test if it'd allow updates assuming the NTP servers are not accessible (maybe there's a logic when it allows those changes).

Anyway, I see your point and I find it reasonable, maybe the timezone setter & getter should accept and return a (pytz?) timezone object?

jeffsf:
I've pulled the access to my DNS-spoofed NTP server from my HS110 (It's on a non-routed VLAN), power cycled it, and will let it thrash through it's list of servers for a while and then re-check to see if the tie can be sent if it can't reach an NTP server.

I'll also look at my HS105, assuming that I can get it to take WiFi credentials through either pyHS100 or tplink-smartplug. While I can get the relay state, it wasn't responding to the "netif" commands for "get_scaninfo" or "set_stainfo"

I'll take a look at `pytz` and see if it makes sense somehow. From what I can tell, there are "only" 109 zones in the TP-Link table, but
```
>>> from pytz import common_timezones
>>> len(common_timezones)
439
```
I've got some roughed out code that uses the captured results of probing a device to be able to convert the index to a fuller description, as well as a human-readable string, as well as dumping the entire list into markdown. It's a lot more than a half-dozen lines of code, so I'll have to figure out your testing framework before I consider a pull request on that code, itself. 

It does let me provide a TIMEZONES.md that I'll add to this once I better understand how the HS110 and HS105 react.

jeffsf:
Very interesting, it might be that you can only set the time _once_ per boot.
```
>>> tp110._query_helper('time', 'set_timezone', {"year":2016, "month":1, "mday":1, "hour":10, "min":10, "sec":10, "index":5})
{}
>>> print(tp110.timezone)
{'index': 5, 'zone_str': '(UTC-08:00) Pacific Standard Time (US & Canada)', 'tz_str': 'PST8', 'dst_offset': 0}
>>> print(tp110.time)
2016-01-01 10:10:28
tp110._query_helper('time', 'set_timezone', {"year":2000, "month":1, "mday":2, "hour":3, "min":4, "sec":5, "index":6})
{}
>>> print(tp110.time)
2016-01-01 10:11:19
tp110._query_helper('time', 'set_timezone', {"year":2017, "month":1, "mday":2, "hour":3, "min":4, "sec":5, "index":6})
{}
>>> print(tp110.time)
2016-01-01 10:12:12
```
Confirmed that, at least on the HS110, you can set the time once on boot, and not again, at not at least within a few minutes. Also that NTP will override manually set time and that manually setting the time does not disable NTP.

kirichkov:
They might be using the time in some manner when communicating with the "cloud". They definitely need that to ensure certificate validity. That can explain why they are limiting the time setting.

jeffsf:
Please put this on hold until I can determine how to make this work for the HS105 as well.
```
>>> tp105._query_helper('time', 'set_timezone', {"year":2001, "month":1, "mday":2, "hour":3, "min":4, "sec":5, "index":6})
{'err_msg': 'invalid argument'}
>>> tp110._query_helper('time', 'set_timezone', {"year":2001, "month":1, "mday":2, "hour":3, "min":4, "sec":5, "index":6})
{}
```

rytilahti:
Interesting findings, thanks for keeping us up to date. Waiting for a more proper solution sounds fair enough.

kirichkov:
@jeffsf do you have any news?

rytilahti:
Adding a dictionary containing the values from https://github.com/GadgetReactor/pyHS100/issues/53#issuecomment-296875129 (keyed with `tz_str`?) and making it use that information for the API call would be an acceptable solution, if someone wants to pursue this.

rytilahti:

Issue #54, opened 2017-04-24
when plug hasn't had lat/long setup, exception is thrown


whorfin:
When lat/long have not been set up, sysinfo shows latitude_i and longitude_i set to 0, and there is
no entry for latitude or longitude

```
Traceback (most recent call last):
  File "/usr/local/bin/pyhs100", line 11, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/pyHS100/cli.py", line 70, in state
    click.echo("Location:     %s" % plug.location)
  File "/usr/local/lib/python2.7/dist-packages/pyHS100/pyHS100.py", line 251, in location
    return {"latitude": info["latitude"],
KeyError: u'latitude'
```

jeffsf:
Can not reproduce
```
>>> tpl = SmartPlug(my_ip_address_as_string)
>>> tpl.location
{'latitude': 0, 'longitude': 0}
```
Confirmed that device is returning `"latitude":0,"longitude":0` in its response, so this result is "correct" from the device's perspective,

Line 251 is in the middle of comments and not executable. Given that and that the current code can't be run with Python older than 3.4, this appears to be an issue with an older version of the code base.

whorfin:
Error was from version installed via pip
[...]

PR #56, opened 2017-04-26
Add check to ensure devices with lat/lon with `_i` suffix are supported (#54)

mlebrun:
Added a .gitignore file as well to prevent committing unnecessary files. I can easily remove if it isn't wanted.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,11 @@
+# Compiled python modules.
+__pycache__/
+
+# Setuptools distribution folder.
+/dist/
+
+# Python egg metadata, regenerated from source files by setuptools.
+/*.egg-info
+/*.egg
+
+/build
--- a/pyHS100/pyHS100.py
+++ b/pyHS100/pyHS100.py
@@ -291,8 +291,19 @@ class SmartDevice(object):
         :rtype: dict
         """
         info = self.sys_info
-        return {"latitude": info["latitude"],
-                "longitude": info["longitude"]}
+        loc = {"latitude": None,
+               "longitude": None}
+
+        if "latitude" in info and "longitude" in info:
+            loc["latitude"] = info["latitude"]
+            loc["longitude"] = info["longitude"]
+        elif "latitude_i" in info and "longitude_i" in info:
+            loc["latitude"] = info["latitude_i"]
+            loc["longitude"] = info["longitude_i"]
+        else:
+            _LOGGER.warning("Unsupported device location.")
+
+        return loc
 
     @property
     def rssi(self):
--- a/pyHS100/tests/fakes.py
+++ b/pyHS100/tests/fakes.py
@@ -45,6 +45,28 @@ emeter_units_support = {"get_realtime": get_realtime_units,
                         "get_monthstat": get_monthstat_units,
                         "get_daystat": get_daystat_units,}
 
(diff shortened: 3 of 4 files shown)


pyHS100/tests/fakes.py
```diff
@@ -46,27 +46,26 @@ def get_daystat_units(obj, x):
...
+                             'feature': 'TIM',
+                             'updating': 0,
+                             'icon_hash': '',
+                             'rssi': 33,
+                             'led_off': 0,
+                             'longitude_i': -12.2,
+                             'latitude_i': 12.2,
+                             'hwId': '60FF6B258734EA6880E186F8C96DDC61',
+                             'fwId': '00000000000000000000000000000000',
+                             'deviceId': '800654F32938FCBA8F7327887A38647617',
```
rytilahti, line 67: This is a white-space fix & change of device id?

.gitignore
```diff
@@ -0,0 +1,11 @@
+# Compiled python modules.
+*.pyc
```
rytilahti: These shouldn't be generated with python3 (which uses ```__pycache__/```)
edit: to add, having a .gitignore file is a good idea otherwise!

pyHS100/pyHS100.py
```diff
@@ -291,8 +291,22 @@ def location(self):
...
         """
         info = self.sys_info
-        return {"latitude": info["latitude"],
-                "longitude": info["longitude"]}
+        loc = {"latitude": 0.0,
+               "longitude": 0.0}
+
+        if "latitude" in info:
+            loc["latitude"] = info["latitude"]
+
```
rytilahti: I would do this like following:
```
if "latitude" in info and "longitude" in info:
    # set both here
elif "latitude_i" in info and "longitude_i" in info:
    # the same
else:
    # log a warning about probably unsupported device?
```

It would also make sense to initialize latitude and longitude to None instead of 0.0, to differentiate between non-existing and well, 0.0/0.0 coordinate.

pyHS100/tests/fakes.py
mlebrun: yea, the device ID was too long for the line, so I figured if the ID isn't specifically being validated as a certainly length and is just most likely a dummy string anyways, I'd just shorten it so Hound would stop complaining. Any non-gitignore change in that commit is likely because Hound was yelling at me.

pyHS100/pyHS100.py
mlebrun: well if you're at 0.0/0.0, you're literally in the middle of the ocean, which I highly doubt is a plausible case, but it could be useful as a visual cue between set vs not set.

.gitignore
mlebrun: will update. I think a couple of times I mixed up `python3` with `python` and `pip`/`pip3` on my dev machine, so likely just a stupid mistake 😅  but fixes incoming.

mlebrun: updating to use `__pycache__/` but let me know if there's any other from this default list from github that you may want included and I can just toss them in: https://github.com/github/gitignore/blob/master/Python.gitignore

mlebrun:
Updated and properly moved the fixes into the correct commits instead of just continuing to amend the last commit

rytilahti:
Looks good to me, thanks for the patch!

PR #58, opened 2017-05-15
[WIP] [SmartDevice] - Enhancement - Adds Various API features

kdschlosser:
I do not personally own one of these devices. I have been working on incorporating the TP-Link smart devices into a piece of software that I am working on. I have not tested this code in any way. If there is someone that has one of these devices and is willing to test it that would be a huge help.

I did my best to stay with the coding style and I added docstings to all of the API methods and properties. I am sure as with everything there will be some changes that have to be made.

I had to port the original code to python 2.7 and this was written in that and I think i have gotten everything converted to python 3 that should be. If i missed something I am sorry for that.

I added classes for the 3 different "schedule" types. these classes are instanes that run inside of the SmartDevice class.

the following commands will return {name: rule} for all of the rules the device has

    SmartDevice.schedule.get()
    SmartDevice.antitheft.get()
    SmartDevice.countdown.get()

I set up the classes to function like a dict so if you want to add/edit

    SmartDevice.schedule['new rule'] = {
        "stime_opt": 0,
        "wday": [0, 1, 0, 1, 1, 0, 0],
        "smin": 1014,
        "enable": 1,
        "repeat": 1,
        "etime_opt": -1,
        "eact": -1,
        "month": 0,
        "sact": 1,
        "year": 0,
        "longitude": 0,
        "day": 0,
        "force": 0,
        "latitude": 0,
        "emin": 0,
    }

the __setitem__ will automatically determine if there is a rule of the same name and if so it will edit that rule. other wise it will make a new rule.

if you want to delete all of the rules you simply delete the schedule

    del(SmartDevice.schedule)

this causes the command to be sent to delete all of the scheduled rules and also will call the smartdevice class to make the schedule instance again

and if you want to delete a single rule

    del(SmartDevice.schedule['rule'])

you can also iterate through the rules with name, rule being yielded
next is implemented with schedules only and will return the next scheduled action to take place

I was thinking about making each rule a class of it's own in the same manner in which the SmartDevice class is done. this way it would be very easy to change anything about the rule. I may still do that if only for my own use that would be fine. but if it is a feature you would like to see done i will gladly submit a PR for it.

Some of the API parameters will need to be tested for purpose and a proper description needs to be added to the docstring 

Also not sure of the return types for some of the API features  
Added API features:

system:
setting device id
setting hardware id
setting location
rebooting
testing uboot
checking config
resetting device to factory defaults

firmware:
setting firmware download url
getting firmware list
downloading firmware
installing firmware

wifi:
scanning access points
connecting to access point

cloud:
getting cloud information
setting cloud server url
connecting to cloud server
unregistering device from the cloud

meter:
getting meter voltage gain and current gain
setting meter voltage gain and current gain
meter calibration

schedule:
getting schedule
getting next scheduled action
editing a schedule item
deleting schedule item
delete schedule for a device
   
countdown - action to perform one time after a given period of time:
adding countdown
editing countdown
deleting countdown
deleting all countdowns

anti-theft - actions to perform when mode is set to away:
get anti-theft list
new anti-theft
edit anti-theft
delete anti-theft
delete all anti-theft

let me know if i should continue to submit PR's for this or not. or if you are not interested in using any of this code.

### DIFF
--- a/pyHS100/antitheft.py
+++ b/pyHS100/antitheft.py
@@ -19,7 +19,7 @@ class AntiTheft(object):
     def __init__(self, _device, _query_helper):
         self._device = _device
         self._query_helper = _query_helper
-    
+
     @property
     def _get_antitheft(self):
         """
@@ -29,12 +29,12 @@ class AntiTheft(object):
         :rtype: list
         :raises SmartPlugException: on error
         """
-        
+
         return self._query_helper(
             "anti_theft",
             "get_rules"
         )
-    
+
     def __iter__(self):
         """
         Iterates all rules.
@@ -43,10 +43,10 @@ class AntiTheft(object):
         :rtype: iterator
         :raises SmartPlugException: on error
         """
-        
+
         for antitheft in self._get_antitheft:
             yield antitheft['name'], antitheft
-    
+
     def __contains__(self, name):
         """
         Checks if rule name exists.
@@ -56,12 +56,12 @@ class AntiTheft(object):
         :rtype: bool
         :raises SmartPlugException: on error
         """
-        
+
         for n, _ in self:
             if n == name:
                 return True
         return False
-    
+
     def __setitem__(self, name, rule):
         """
         Edits or Adds a new rule.
@@ -94,13 +94,13 @@ class AntiTheft(object):
         :rtype: None
         :raises SmartPlugException: on error
         """
-        
+
         if name in ('_device', '_query_helper'):
             object.__setattr__(self, name, rule)
             return
-        
+
         rule['name'] = name
-        
+
         if name in self:
             rule['id'] = self[name]['id']
             self._query_helper(
@@ -114,7 +114,7 @@ class AntiTheft(object):
                 "add_rule",
                 rule
             )
-    
+
     def __delitem__(self, name):
         """
         Deletes rule.
@@ -125,7 +125,7 @@ class AntiTheft(object):
         :raises NoAntiTheftFound: If rule not found
         :raises SmartPlugException: on error
         """
-        
+
         for n, a in self:
             if n == name:
                 return self._query_helper(
@@ -134,7 +134,7 @@ class AntiTheft(object):
                     {"id": a['id']}
                 )
         raise NoAntiTheftFound(name)
-    
+
     def get(self, name=None, *args):
         """
         Gets rule
@@ -147,14 +147,14 @@ class AntiTheft(object):
         :raises NoAntiTheftFound: If rule not found
         :raises SmartPlugException: on error
         """
-        
+
         if name is None:
             return dict(list(antitheft for antitheft in self))
         elif args and name not in self:
             return args[0]
         else:
             return self[name]
-    
+
     def __getitem__(self, name):
         """
         Gets rule
@@ -165,16 +165,16 @@ class AntiTheft(object):
         :raises NoAntiTheftFound: If rule not found
         :raises SmartPlugException: on error
         """
-        
+
         if name in self.__dict__:
             return self.__dict__[name]
-        
+
         for n, a in self:
             if n == name:
                 return a
-        
+
         raise NoAntiTheftFound(name)
-    
+
     def __del__(self):
         """
         Deletes all rules.
@@ -183,18 +183,17 @@ class AntiTheft(object):
         :rtype: None
         :raises SmartPlugException: on error
         """
-        
+
         self._query_helper(
             "anti_theft",
             "delete_all_rules"
         )
         self._device.new_antitheft()
-    
-    
+
+
 class NoAntiTheftFound(Exception):
-    
     def __init__(self, name):
         self.msg = 'There is no antitheft by the name %s.' % name
-        
+
     def __str__(self):
         return self.msg
--- a/pyHS100/antitheft.py
+++ b/pyHS100/antitheft.py
@@ -1,0 +1,200 @@
+"""
+pyHS100
+Python library supporting TP-Link Smart Plugs/Switches (HS100/HS110/Hs200).
+
+The communication protocol was reverse engineered by Lubomir Stroetmann and
+Tobias Esser in 'Reverse Engineering the TP-Link HS110':
+https://www.softscheck.com/en/reverse-engineering-tp-link-hs110/
+
+This library reuses codes and concepts of the TP-Link WiFi SmartPlug Client
+at https://github.com/softScheck/tplink-smartplug, developed by Lubomir
+Stroetmann which is licensed under the Apache License, Version 2.0.
+
+You may obtain a copy of the license at
+http://www.apache.org/licenses/LICENSE-2.0
+"""
+
+
+class AntiTheft(object):
+    def __init__(self, _device, _query_helper):
+        self._device = _device
+        self._query_helper = _query_helper
+    
+    @property
+    def _get_antitheft(self):
+        """
+        Gets all rules.
+
+        :return: list rules
+        :rtype: list
+        :raises SmartPlugException: on error
+        """
+        
+        return self._query_helper(
+            "anti_theft",
+            "get_rules"
+        )
+    
+    def __iter__(self):
+        """
+        Iterates all rules.
+
+        :return: iterator (name, rule)
+        :rtype: iterator
+        :raises SmartPlugException: on error
+        """
+        
+        for antitheft in self._get_antitheft:
+            yield antitheft['name'], antitheft
+    
+    def __contains__(self, name):
+        """
+        Checks if rule name exists.
+
+        :param str name: name of the rule
+        :return: bool True if name exists else False
+        :rtype: bool
+        :raises SmartPlugException: on error
+        """
+        
+        for n, _ in self:
+            if n == name:
+                return True
+        return False
+    
+    def __setitem__(self, name, rule):
+        """
+        Edits or Adds a new rule.
+
+        If a rule exists with the same name it will be updated.
+        Otherwise it will be added.
+
+        Value must be a dict with the following key, value pairs
+        "stime_opt":int(0)
+        "wday":list(0, 1, 0, 1, 1, 0, 0), 0 or a 1 for each day of the week
+                                          1 representing a scheduled day
+        "smin":int(1014)
+        "enable":int(1), 0 or a 1, 1 being enabled
+        "frequency": int(5)
+        "repeat": int(1), 0 or a 1, 1 to repeat the schedule
+        "etime_opt": int(-1)
+        "duration": int(2)
+        "lastfor": int(1)
+        "month": int(0), 0 for all months
+        "year": int(0), 0 for all years
+        "longitude": int/float(0), 0 for all locations
+        "day": int(0), 0 for all days of the month
+        "force": int(0), 0 or 1, 1 to force schedule
+        "latitude": int/float(0), 0 for all locations
+        "emin": int(0)
+
+        :param str name: Name of the rule
+        :param dict rule: Parameters for the rule (see description)
+        :return: None
+        :rtype: None
+        :raises SmartPlugException: on error
+        """
+        
+        if name in ('_device', '_query_helper'):
+            object.__setattr__(self, name, rule)
+            return
+        
+        rule['name'] = name
+        
+        if name in self:
+            rule['id'] = self[name]['id']
+            self._query_helper(
+                "anti_theft",
+                "edit_rule",
+                rule
+            )
+        else:
+            self._query_helper(
+                "anti_theft",
+                "add_rule",
+                rule
+            )
+    
+    def __delitem__(self, name):
+        """
+        Deletes rule.
+
+        :param str name: Name of the rule
+        :return: None
+        :rtype: None
+        :raises NoAntiTheftFound: If rule not found
+        :raises SmartPlugException: on error
+        """
+        
+        for n, a in self:
+            if n == name:
+                return self._query_helper(
+                    "anti_theft",
+                    "delete_rule",
+                    {"id": a['id']}
+                )
+        raise NoAntiTheftFound(name)
+    
+    def get(self, name=None, *args):
+        """
+        Gets rule
+
+        :param str name: Name of the rule.
+        :param None name: If None returns all rules.
+        :param args: Used for default return value.
+        :return: dict name: All rules if item is None else rule
+        :rtype: dict
+        :raises NoAntiTheftFound: If rule not found
+        :raises SmartPlugException: on error
+        """
+        
+        if name is None:
+            return dict(list(antitheft for antitheft in self))
+        elif args and name not in self:
+            return args[0]
+        else:
+            return self[name]
+    
+    def __getitem__(self, name):
+        """
+        Gets rule
+
+        :param str name: Name of the rule.
+        :return: dict rule
+        :rtype: dict
+        :raises NoAntiTheftFound: If rule not found
+        :raises SmartPlugException: on error
+        """
+        
+        if name in self.__dict__:
+            return self.__dict__[name]
+        
+        for n, a in self:
+            if n == name:
+                return a
+        
+        raise NoAntiTheftFound(name)
+    
+    def __del__(self):
+        """
+        Deletes all rules.
+
+        :return: None
+        :rtype: None
+        :raises SmartPlugException: on error
+        """
+        
+        self._query_helper(
+            "anti_theft",
+            "delete_all_rules"
+        )
+        self._device.new_antitheft()
+    
+    
+class NoAntiTheftFound(Exception):
+    
+    def __init__(self, name):
+        self.msg = 'There is no antitheft by the name %s.' % name
+        
+    def __str__(self):
+        return self.msg
--- a/pyHS100/pyHS100.py
+++ b/pyHS100/pyHS100.py
@@ -19,13 +19,16 @@ import socket
 import enum
 
 from .protocol import TPLinkSmartHomeProtocol
-from .schedule import Schedule, NoScheduleFound
-from .countdown import Countdown, NoCountdownFound
-from .antitheft import AntiTheft, NoAntiTheftFound
+from . import schedule
+from . import countdown
+from . import antitheft
+
 
 _LOGGER = logging.getLogger(__name__)
 
-
+NoScheduleFound = schedule.NoScheduleFound
+NoCountdownFound = countdown.NoCountdownFound
+NoAntiTheftFound = antitheft.NoAntiTheftFound
 class SmartPlugException(Exception):
     """
     SmartPlugException gets raised for errors reported by the plug.
@@ -60,9 +63,9 @@ class SmartDevice(object):
             protocol = TPLinkSmartHomeProtocol()
         self.protocol = protocol
 
-        self.schedule = Schedule(self, self._query_helper)
-        self.countdown = Countdown(self, self._query_helper)
-        self.antitheft = AntiTheft(self, self._query_helper)
+        self.schedule = schedule.Schedule(self, self._query_helper)
+        self.countdown = countdown.Countdown(self, self._query_helper)
+        self.antitheft = antitheft.AntiTheft(self, self._query_helper)
 
     def new_schedule(self):
         """
@@ -72,7 +75,7 @@ class SmartDevice(object):
         :rtype: None
         """
 
-        self.schedule = Schedule(self, self._query_helper)
+        self.schedule = schedule.Schedule(self, self._query_helper)
 
     def new_countdown(self):
         """
@@ -82,7 +85,7 @@ class SmartDevice(object):
         :rtype: None
         """
 
-        self.countdown = Countdown(self, self._query_helper)
+        self.countdown = countdown.Countdown(self, self._query_helper)
 
     def new_antitheft(self):
         """
@@ -91,7 +94,7 @@ class SmartDevice(object):
         :rtype: None
         """
 
-        self.antitheft = AntiTheft(self, self._query_helper)
+        self.antitheft = antitheft.AntiTheft(self, self._query_helper)
 
     def _query_helper(self, target, cmd, arg=None):
         """
--- a/pyHS100/pyHS100.py
+++ b/pyHS100/pyHS100.py
@@ -71,7 +71,7 @@ class SmartDevice(object):
         :return: None
         :rtype: None
         """
-    
+
         self.schedule = Schedule(self, self._query_helper)
 
     def new_countdown(self):
@@ -81,7 +81,7 @@ class SmartDevice(object):
         :return: None
         :rtype: None
         """
-    
+
         self.countdown = Countdown(self, self._query_helper)
 
     def new_antitheft(self):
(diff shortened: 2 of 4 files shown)


kdschlosser:
sorry about the commits. I do like my code to pass all checks 100%

kirichkov:
Thanks for your contribution but you should also supply tests with it. Otherwise it might break at any moment. This is quite a lot of code that without tests will require to someone periodically check it whether it works or not using a physical device.

kdschlosser:
Yesssir. That is one of the reasons as to why I have issued a PR for it. You will have far more foot traffic for this and someone may be willing to test it out.

kdschlosser:
I m adding this code to the GUI front end I made which is going to take me a little bit to finish that end of it up. there is a whole lot of code involved for that end of it.

kirichkov:
I'm personally opposed to admitting code that is not tested, but I'd agree if @rytilahti and @mweinelt agree to accept it, so I'll leave that open for them to look on and share opinion.

kdschlosser:
oh lord knows i do not want anyone to just accept it... it needs to be tested. I agree with you 100% on that. Doing a PR for for the code give it more exposure to people that have the ability to test it. I will get the code tested soon enough. just that coding all of the gui based stuff is a pretty large undertaking. and I do have some things that i have to do before i can finish that up. so if you are not opposed to just letting the PR sit for a week or so. I am hoping i should be able to get some information back to ya by then.

kirichkov:
I am talking about writing automated tests, not physical tests with a real device. Please take a look at the files in the [tests](https://github.com/GadgetReactor/pyHS100/tree/master/pyHS100/tests) directory.

For every functionality there is a number of "tests" - these are functions that are automatically run and ensure that when something is changed the API does not break.

This is to make sure that when someone else modifies your code and opens a pull request the new code behaves the same as when you wrote it. Let me oversimplify it: When supplying a parameter of 1 to your method, that has code changed, it still returns 2 and not 3 after the code change :-)

P.S. I do not mind keeping the pull request open. Just check once in a while whether you got pinged here if you need more than a week. I'll also add a [WIP] to the title to be clear.

kdschlosser:
I got ya. so you want me to add the means to test the methods. That's not a problem. once i can verify that the methods actually work like they are intended and get some of the parameters sorted out because the API is not clear on what some of them actually do. I think i am going to have to bite the bullet and buy one of these things.

kirichkov:
The major problem is that some of the functionality is not available in the HS-105, but none of us who are maintaining this repo have a HS-105, so we can't test it even if you wanted to. Perhaps some of the users of your software can land a hand and do the testing for the different module?

Keep in mind the same protocol is used not just for the power plugs, but also for in-wall switches and the TP-Link light bulbs. The switch and the bulbs are not sold in Europe last time I checked, so it's not as straightforward as you think to properly test it manually.

kdschlosser:
I have not come across someone with a light bulb here in the US. But these things are cheap so I will order some of them simply for having something to test with to make sure everything works as intended. some of the things I am going to have to play around with to much to try and explain to someone how to test it. as an example. connecting the wifi. it's a number you pass to tell it what encryption to use. don't know what numbers do what and what the numbers are for that matter

rytilahti:
Hi Kevin and thanks for the PR! I'm going to go it briefly through and leave some comments behind, however, if I may ask, could you submit a separate PR for SmartDevice's new features and another for the scheduling ones (countdown, antitheft, schedule)?

The new features in SmartDevice seems mostly straightforward and I see no problems adding those without larger changes, however as @kirichkov mentioned it'd be nice to have unit tests for those too.

To me the API for scheduling feels a bit cumbersome and probably error-prone. Considering those are all related in a way or another to scheduling/automation, do you think they could share parts of the logic, and just expose different types as enums?

pyHS100/antitheft.py
```diff
@@ -0,0 +1,199 @@
...
+        "duration": int(2)
+        "lastfor": int(1)
+        "month": int(0), 0 for all months
+        "year": int(0), 0 for all years
+        "longitude": int/float(0), 0 for all locations
+        "day": int(0), 0 for all days of the month
+        "force": int(0), 0 or 1, 1 to force schedule
+        "latitude": int/float(0), 0 for all locations
+        "emin": int(0)
+
```
rytilahti, line 90: Would it make sense to have a wrapper class for this information? In any case I think it would make sense to use e.g. voluptuous (https://github.com/yadutaf/voluptuous) to enforce correct input values.

pyHS100/antitheft.py
```diff
@@ -0,0 +1,199 @@
...
+        """
+        Iterates all rules.
+
+        :return: iterator (name, rule)
+        :rtype: iterator
+        :raises SmartPlugException: on error
+        """
+
+        for antitheft in self._get_antitheft:
+            yield antitheft['name'], antitheft
```
rytilahti, line 48: Would it make sense just to yield from _get_antitheft and let the user to extract the name from the resulting dict? It would make the code more approachable.

pyHS100/antitheft.py
```diff
@@ -0,0 +1,199 @@
...
+This library reuses codes and concepts of the TP-Link WiFi SmartPlug Client
+at https://github.com/softScheck/tplink-smartplug, developed by Lubomir
+Stroetmann which is licensed under the Apache License, Version 2.0.
+
+You may obtain a copy of the license at
+http://www.apache.org/licenses/LICENSE-2.0
+"""
+
+
+class AntiTheft(object):
```
rytilahti, line 18: Python 3 doesn't require explicit inheritance from object, so
```class AntiTheft:``` would work just fine.

File: pyHS100/antitheft.py [hunk 4]
```diff
@@ -0,0 +1,199 @@
...
+    @property
+    def _get_antitheft(self):
+        """
+        Gets all rules.
+
+        :return: list rules
+        :rtype: list
+        :raises SmartPlugException: on error
+        """
+
```
rytilahti, line 32: No need for an empty line between doctstring and the code.

pyHS100/antitheft.py
```diff
@@ -0,0 +1,199 @@
...
+        self._query_helper(
+            "anti_theft",
+            "delete_all_rules"
+        )
+        self._device.new_antitheft()
+
+
+class NoAntiTheftFound(Exception):
```
rytilahti, line 194: In my opinion the exception should inherit SmartPlugException to allow easier exception handling.

pyHS100/antitheft.py
```diff
@@ -0,0 +1,199 @@
...
+    def __init__(self, name):
+        self.msg = 'There is no antitheft by the name %s.' % name
```
rytilahti, line 196: Is this really necessary?
```
class NoAntiTheftFound(SmartPlugException):
    pass
```
and then just:
```
raise NoAntiTheftFound("No antitheft rule by the name %s was found" % name)
```

pyHS100/antitheft.py
```diff
@@ -0,0 +1,199 @@
...
+        :param str name: Name of the rule
+        :param dict rule: Parameters for the rule (see description)
+        :return: None
+        :rtype: None
+        :raises SmartPlugException: on error
+        """
+
+        if name in ('_device', '_query_helper'):
+            object.__setattr__(self, name, rule)
+            return
```
rytilahti, line 100: This feels a bit fragile. Is there some good reason to expose these internals?

pyHS100/pyHS100.py
```diff
@@ -19,10 +19,16 @@
...
+from . import countdown
+from . import antitheft
 
-_LOGGER = logging.getLogger(__name__)
 
+_LOGGER = logging.getLogger(__name__)
 
+NoScheduleFound = schedule.NoScheduleFound
+NoCountdownFound = countdown.NoCountdownFound
+NoAntiTheftFound = antitheft.NoAntiTheftFound
```
rytilahti, line 31: These are for the exceptions? Maybe we should have a separate exception file containing all our custom exceptions instead of splitting them into different submodules and importing them like this.

pyHS100/pyHS100.py
```diff
@@ -57,6 +63,39 @@ def __init__(self, ip_address, protocol=None):
...
+
+    def new_antitheft(self):
+        """
+        Used by AntiTheft instance when it gets deleted.
+        :return: None
+        :rtype: None
+        """
+
+        self.antitheft = antitheft.AntiTheft(self, self._query_helper)
+
```
rytilahti, line 98: How/why does those modules need to do this? This entangles the logic of those submodules to this class (instead of having it other way around) and it doesn't feel like a good idea to do.

pyHS100/pyHS100.py
```diff
@@ -305,6 +344,39 @@ def location(self):
...
+        :raises SmartPlugException: on error
+        """
+
+        latitude, longitude = location
+
+        if latitude < 0:
+            latitude = -latitude
+        if longitude < 0:
+            longitude = -longitude
+
```
rytilahti, line 362: Shouldn't it be possible to have negative lats and lons?

pyHS100/pyHS100.py
```diff
@@ -480,3 +552,280 @@ def state_information(self):
...
+    @property
+    def device_id(self):
+        """
+        Gets the device id.
+
+        :return: Device id
+        :rtype: str
+        :raises SmartPlugException: on error
+        """
+
```
rytilahti, line 585: Empty line.

pyHS100/pyHS100.py
```diff
@@ -480,3 +552,280 @@ def state_information(self):
...
+        self._query_helper("system", "check_new_config")
+
+    def scan_ap(self, refresh=1):
+        """
+        Scans for access points.
+
+        :param int refresh: Refresh interval in seconds.
```
rytilahti, line 686: Will this keep on scanning on interval of refresh until it gets canceled somehow? By connecting it to an access point? The behaviour should be documented here I think :-)

pyHS100/pyHS100.py
```diff
@@ -480,3 +552,280 @@ def state_information(self):
...
+        )
+
+    # TODO: map out the key_types
+    def connect_ap(self, ssid='', password='', key_type=0):
+        """
+        Connects to an access point.
+
+        :param str ssid: Access point ssid.
+        :param str password: Access point password.
+        :param int key_type: Access point encryption type.
```
rytilahti, line 705: This should be an IntEnum, or the acceptable values should at least be commented.

rytilahti (commented):
Some initial comments on the code. I didn't mark all extra newlines between docstrings and code, but those should also be cleaned :-)

pyHS100/antitheft.py
kdschlosser: habit LOL

pyHS100/antitheft.py
kdschlosser: can clean up the newlines after the doc strings not a problem.

pyHS100/antitheft.py
kdschlosser: So what takes place is when a user wants to create or modify a rule they can do the following this is truncated for sake of the explanation 

    device_instance.antitheft['some rule'] = {example rules}

and what __setitem__ does is it will check if the key 'some_rule' already exists in self, __contains__ which calls the iterator. to return a name, rule pairing checks to see if the name matches. and return true.

then if it is true the rule is retrieved and the id of the rule is copied to the new rule vales along with the name of the rule and is sent to the device as a modify request. otherwise it is sent off as a new rule request.

it wasn't specifically designed for the user but seeing how it can be i think it would also make it simpler as an example

    name_piece = 'some':
    for name, rule in device_instance.antitheft:
        if name.startswith(name_piece):
            rule['parameter'] = 1
            device_instance.antitheft[name] = rule
            break

pyHS100/antitheft.py
kdschlosser: i was going to create a base class for handling all 3 of the "schedule type classes" I just wasn't sure which way you would have wanted the coding done. it's easier to create the base class from this point then it would have been to split it into 3 different classes.

pyHS100/antitheft.py
kdschlosser: that i would have done but because SmartPlugException class is in pyHS100.py which in the same file where the 3 new classes are imported into so there was now way i would have been able to import it into any of those files to create a subclass of it due to circular imports. and i couldn't have created the subclasses in the pyHS100.py file either due to the same thing. so unless there is a separate exception file created with SmartPlugException in it then i am not sure how to go about subclassing it.

pyHS100/pyHS100.py
kdschlosser: the reason for this being done is because if you wanted to delete all scheduled/countdown/antitheft rules you would simply delete the instance from the device.

    del device_instance.antitheft

would cause the following code to run removing all antitheft rules form the device. but the antitheft instance still needs to exist so if the user wanted to make new rules

        def __del__(self):
            """
            Deletes all rules.

            :return: None
            :rtype: None
            :raises SmartPlugException: on error
            """

            self._query_helper(
                "anti_theft",
                "delete_all_rules"
            )
            self._device.new_antitheft()

pyHS100/pyHS100.py
kdschlosser: my goof on that one. my intention was to make the positive ones negative. tho have not done to much research into lat and lon. but i do know that some times my lat and lon have come up either positive or negative and only the negative ones are correct. I went based off of that. is there even such a thing as a positive lat and lon? I do not know

pyHS100/pyHS100.py
kdschlosser: I do not know the answer to that. and there is no cancel that i have seen. my assumption is that it will continually scan. I would need a device to test it with.

pyHS100/pyHS100.py
kdschlosser: I do not know what the acceptable numbers are. there is no documentation on this. this would be one of the many reasons why i would need an actual device to test. there is no existing documentation of what numbers are what security type

pyHS100/pyHS100.py
kdschlosser: we would need to create a separate exception file and also have the SmartPlugException class in it in order to be able to subclass it.

pyHS100/antitheft.py
kdschlosser: would have done this. no way to subclass the SmartPlugException due to circular imports

pyHS100/antitheft.py
kdschlosser: you are correct. this is a piece of code that i should have removed. i was originally going to make it so that the user could access a specific rule using __setattr__ and decided against it due to special characters and forgot to delete this code.

pyHS100/antitheft.py
rytilahti: I think it makes sense to have a separate file for exceptions/commonalities such as exceptions & enums.

pyHS100/antitheft.py
rytilahti: Indeed, see above. I think there's a need for some refactoring to be done in pyHS100.py.

pyHS100/antitheft.py
kdschlosser: and also i ported this code back to python 3 i have ported the original code over to python 2 and added everything there.

pyHS100/pyHS100.py
rytilahti: The deletion semantic (and overriding __del__) is something not very pythonic, one usually does not see ```del``` calls in the code if that's not for a very specific use case. I'd say that it'd be better to have a regular method for deletion [1], then it's explicitly clear what is being done.

[1] If Rules are to be a separate class like discussed above, maybe just delete/remove method on those? Deleting all rules can easily be done by looping all the rules and deleting them.

pyHS100/pyHS100.py
rytilahti: Yeah, http://www.mapsofworld.com/images2008/world-map-with-latitude-and-longitude.jpg negatives are for north and west :-)

pyHS100/pyHS100.py
rytilahti: Ok, I was just curious.

pyHS100/antitheft.py
kdschlosser: i didn't want to go about moving all kinds of things around. this is more of a proof of concept that is obviously going to be changed but is also going to require changes to the original code. it. not my place to to do that. 😄 So i thought it best to leave it to up to you.

pyHS100/pyHS100.py
rytilahti: Ok, not a problem. Maybe a good idea to leave a TODO behind for it. I can't play around with it at the moment as my plug is "hosting" my computer setup. Maybe I'll check that out when I unplug it at some point.

pyHS100/antitheft.py
rytilahti: Hm, ok. I didn't see this particular piece to be handled below?

pyHS100/pyHS100.py
kdschlosser: ok so that needs to be removed and the user does whatever. and if it is wrong then that's on them. LOL

pyHS100/antitheft.py
rytilahti: Ok :+1:

pyHS100/pyHS100.py
kdschlosser: i'm curious as well. there are quite a few things that i do not know the specifics to.

pyHS100/antitheft.py
kdschlosser: I do have a better idea on how to handle this. and i think it would be far easier to understand if i simply code it up then to try and explain it. should only take me an hour or so.

pyHS100/pyHS100.py
kdschlosser: funny.. I have the TODO in my original code. not sure how it didn't make it into this one. I will make sure those make it to the next revision of this.

kdschlosser:
I hope this cleared up any questions. But it also gave some direction where to go with it.
do you want me to take care of putting the SmartPlugException into it's own file and issue a PR for it? or do you want to take care of that end of it. and then i can go from there? 

I will split the PR into 2 parts. one for the SmartDevice alterations without any of the scheduling things added. and then I will issue a PR for each of the different scheduling types. is that acceptable?

pyHS100/pyHS100.py
kdschlosser: up to you. i can also create a specific method for deleting all rules as well without the need to loop.

kdschlosser:
@kirichkov mentioned the lack of access to the bulbs. here is the data from an LB100  Hope this helps

    {
        u'ip': '192.168.1.100',
        u'port': 9999,
        u'sys_info': {
            u'emeter': {
                u'err_code': -2001,
                u'err_msg': u'Module not support'
            },
            u'system': {
                u'get_sysinfo': {
                    u'active_mode': u'none',
                    u'alias': u'New Light',
                    u'ctrl_protocols': {
                        u'name': u'Linkie',
                        u'version': u'1.0'
                    },
                    u'description': u'Smart Wi-Fi LED Bulb with Dimmable Light',
                    u'dev_state': u'normal',
                    u'deviceId': u'8012996ED1F8DA43EFFD58B62BEC5ADE18192F88',
                    u'disco_ver': u'1.0',
                    u'err_code': 0,
                    u'heapsize': 340808,
                    u'hwId': u'111E35908497A05512E259BB76801E10',
                    u'hw_ver': u'1.0',
                    u'is_color': 0,
                    u'is_dimmable': 1,
                    u'is_factory': False,
                    u'is_variable_color_temp': 0,
                    u'light_state': {
                        u'dft_on_state': {
                            u'brightness': 50,
                            u'color_temp': 2700,
                            u'hue': 0,
                            u'mode': u'normal',
                            u'saturation': 0
                        },
                        u'on_off': 0
                    },
                    u'mic_mac': u'50C7BF3393F1',
                    u'mic_type': u'IOT.SMARTBULB',
                    u'model': u'LB100(US)',
                    u'oemId': u'264E4E97B2D2B086F289AC1F00B90679',
                    u'preferred_state': [
                        {
                            u'brightness': 100,
                            u'color_temp': 2700,
                            u'hue': 0,
                            u'index': 0,
                            u'saturation': 0
                        },
                        {
                            u'brightness': 75,
                            u'color_temp': 2700,
                            u'hue': 0,
                            u'index': 1,
                            u'saturation': 0
                        },
                        {
                            u'brightness': 25,
                            u'color_temp': 2700,
                            u'hue': 0,
                            u'index': 2,
                            u'saturation': 0
                        },
                        {
                            u'brightness': 1,
                            u'color_temp': 2700,
                            u'hue': 0,
                            u'index': 3,
                            u'saturation': 0
                        }
                    ],
                    u'rssi': -54,
                    u'sw_ver': u'1.2.3 Build 170123 Rel.100146'
                }
            }
        }
    }

pyHS100/antitheft.py
rytilahti: Done in #61.

rytilahti:
I created a PR #61 to extract SmartPlugException (to be called SmartDeviceException) from its class. After it gets merged I think it makes sense for you to rebase your patches on the new master branch.

There's a test data for LB130 in tests/fakes.py, looks like this is similar to it but without color leds and with no color temperature support? Maybe it's worth to add this as another bulb profile to fakes and adjust tests to test against both bulbs?

kdschlosser:
I am still working on reading through the tests portion. but if we can get the output for each device model then that would be the way to go to make sure all of the bases are covered with all the devices.

kdschlosser:
now the really odd thing is that this model does not support time_zone either

pyHS100/antitheft.py
kirichkov: Totally off topic. Would you be interested in supporting a python 2 branch? We've dropped the support requirement, but if you're going to support python 2 on a fork, you might as well do it here in this repo to keep all code in one place.

rytilahti: Instead of resurrecting a python2 branch, would it make more sense to convert the downstream code to use python3, if it isn't too great effort?

Anyway, having a proper python2 support (contained in compat.py, similar to https://github.com/requests/requests/blob/master/requests/compat.py) would also be fine.

kirichkov:
LB100 & LB110 are dimmable whites (~probably one is warm white the other cool white~ they should be the same just different power/lumens), 120 is white with adjustable hue, and 130 is full-color bulb. I guess that's why they advertise different features.

I'm basing this on the information on tp-link's site:
http://www.tp-link.com/us/home-networking/smart-home/smart-bulbs

kdschlosser:
There is a slew of differences between the API for the Bulb and the plugs. 

There is no "updating" in sys_info. The mac is actually mic_mac. The type is mic_type there is no "fwId" in sys_info after each query the LB100 just drops connection after it replies so the closing of the socket fails. And there are a few others I can't remember off the top of my head. So I am thinking  it might be a good idea to create a NotSupportedError and also subclass a base class that represents the Bulbs for each of the model numbers. There are not to many models of the various devices available. I think it would make for cleaner code to "discover" the device type and apply the proper class. This way if there is a change like the mac and mic_mac the method can be overridden so we will be able to specifically identify what device/model actually uses which version of the API. But this may also provide a means of seeing if the API changes are in fact not keyed to models but maybe to firmware. Also I have figured out the parameters In the schedule API. I am hoping to be sent to sort out the WiFi as well. I have a beta tester that is able to test these things out for me.

kdschlosser:
And the LB100 has no emeter where as the LB110 does

rytilahti:

PR #59, opened 2017-05-26
[SmartDevice] - Bugfix - Adds LB100 support for getting and setting the mac address

kdschlosser:
LB100 devices instead of using 'mac' as a keyword they use 'mic_mac' so checking for the keyword has been added to the getter and setter

### DIFF
--- a/pyHS100/pyHS100.py
+++ b/pyHS100/pyHS100.py
@@ -323,9 +323,8 @@ class SmartDevice(object):
         :return: mac address in hexadecimal with colons, e.g. 01:23:45:67:89:ab
         :rtype: str
         """
-
         info = self.sys_info
-        
+
         if 'mac' in info:
             return info["mac"]
         elif 'mic_mac' in info:
--- a/pyHS100/pyHS100.py
+++ b/pyHS100/pyHS100.py
@@ -323,7 +323,16 @@ class SmartDevice(object):
         :return: mac address in hexadecimal with colons, e.g. 01:23:45:67:89:ab
         :rtype: str
         """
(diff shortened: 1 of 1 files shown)


pyHS100/pyHS100.py
```diff
@@ -333,7 +341,18 @@ def mac(self, mac):
...
-        self._query_helper("system", "set_mac_addr", {"mac": mac})
+        info = self.sys_info
+
+        if 'mac' in info:
+            self._query_helper("system", "set_mac_addr", {"mac": mac})
+        elif 'mic_mac' in info:
+            self._query_helper(
+                "system",
+                "set_mac_addr",
+                {"mic_mac": mac.replace(':', '')}
```
rytilahti, line 352: Is replacing just necessary for "mic_mac" type? As the calls for both of these are same, I'd handle setting "payload" variable with correct content in the if/else block, and just call query helper after that.

kdschlosser: well i noticed that the mic_mac does not have :'s in the mac address in the first PR i did i posted the actual data that came from a user that has an LB100 and in there the mic_mac does not have :'s

and it can be set up for a single call but this is an if, elif. so in the event there is a 3rd kind wandering around it will not try to set it incorrectly. it would be more code to check if there is payload data before sending the query. however you want to set it up is up to you

rytilahti: I think your solution is good enough, could you rebase and let's get it merged then :)
edit: to add, I used parts of your code already to fix `mac()`, merged in #71.

rytilahti:
Closing as stale.

PR #70, opened 2017-08-03
add @pass_dev to hsv, adjust ranges

rytilahti:
Looks like no one tested the bulb support..

### DIFF
--- a/pyHS100/cli.py
+++ b/pyHS100/cli.py
@@ -136,9 +136,9 @@ def temperature(dev, value):
 
 
 @cli.command()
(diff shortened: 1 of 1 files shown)


kirichkov:
If this is ready to be merged I'd suggest "Squash and merge".

Thanks to whoever is doing the actual testing with the physical device 👍

danodemano:
@kirichkov - that would be me.  :)  With the help of @rytilahti as I only know enough Python to be dangerous.  We've had two solid evenings of testing/conversations via Discord.

PR #71, opened 2017-08-03
Discover refactoring, enhancements to the cli tool

rytilahti:
* Discover tries to detect the type of the device from sysinfo response
* Discover.discover() returns an IP address keyed dictionary,
  values are initialized instances of the automatically detected device type.

* When no IP is given, autodetect all supported devices and print out their states
* When only IP but no type is given, autodetect type and make a call based on that information.
  * One can define --bulb or --plug to skip the detection.

* renamed pyHS100.py -> smartdevice.py

### DIFF
--- a/pyHS100/__init__.py
+++ b/pyHS100/__init__.py
@@ -13,8 +13,9 @@ Module-specific errors are raised as `SmartDeviceException` and are expected
 to be handled by the user of the library.
 """
 # flake8: noqa
+from .smartdevice import SmartDevice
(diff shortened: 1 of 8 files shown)


rytilahti:
This paves the path to automatically detecting supported devices on homeassistant, I would encourage everyone having any devices testing the `pyhs100` tool to see if it behaves as expected when having 1) multiple devices, 2) different types of devices, 3) bulbs in general.

kirichkov (approved):
Looks good to me!

Issue #67, opened 2017-06-30
Query State failed on HS100


Semant1ka:
Hi,
I've run the utility with my smart plug and got this exception on query state.
In order to make discovery working, I've changed target field in TPLinkSmartHomeProtocol.discover to my custom wlan0 broadcast network value.

== My Smart Plug - HS100(EU) ==
Device state: ON
LED state: True
On since: 2017-06-16 09:35:45.934023
== Generic information ==
Time:         1970-01-14 23:19:24
Hardware:     1.0
Software:     1.0.8 Build 151101 Rel.24452
Traceback (most recent call last):
  File "/usr/local/bin/pyhs100", line 9, in <module>
    load_entry_point('pyHS100==0.2.4.2', 'console_scripts', 'pyhs100')()
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/usr/local/lib/python3.4/dist-packages/pyHS100-0.2.4.2-py3.4.egg/pyHS100/cli.py", line 80, in state
    click.echo("MAC (rssi):   %s (%s)" % (dev.mac, dev.rssi))
  File "/usr/local/lib/python3.4/dist-packages/pyHS100-0.2.4.2-py3.4.egg/pyHS100/pyHS100.py", line 302, in rssi
    return self.sys_info["rssi"]
KeyError: 'rssi'

[...]

PR #74, opened 2017-08-04
add hs100 example to fakes.py, thanks to Semant1ka on #67

rytilahti:


rytilahti:
As this does not affect the functionality, I'll just proceed with merge.

PR #75, opened 2017-08-05
Add LB110 sysinfo

gollo:


rytilahti:
Thanks, looks good. As it is just a fake data file I think we can ignore those hound warnings, that can be fixed later if wanted.

PR #72, opened 2017-08-04
wrap sysinfo to defaultdict to return None

rytilahti:
Wraps sysinfo with defaultdict returning None, makes it easier not to fail hard on missing values on different types of devices. Fixes #67

PR #76, opened 2017-08-05
Extend tests

rytilahti:
Run the tests against all known system infos (HS100, HS105, HS110, HS200, LB110, LB130). The error with 'rssi' will be fixed when #72 gets merged.

Also change the API for meter-based functionalities, they will now return None instead of False when the device has no emeter.


rytilahti:
I think those hound warnings on fakes.py can be ignored.

kirichkov (approved):
Looks good. I also think one shouldn't spend more time fixing styling errors than syntax errors ;-)

PR #77, opened 2017-08-05
Prevent failure if device's sysinfo does not have a "feature" attribute.

gollo:


kirichkov:
Thanks for the PR!

I think this should already be resolved for all attributes by #72

rytilahti:
@kirichkov #72 makes the access to return None and calling split() on it will cause an AttributeError, so this PR is still needed.

Thanks @gollo, merging.

PR #81, opened 2017-08-08
make tox run pytest-cov, add coveralls

rytilahti:
Adds coverage report for unit tests, makes it easier to see what parts of code is tested and what is not :-)


kirichkov:
This seems good to go. Merge unless you think 66% is not representative :-)

rytilahti:
Hopefully it'll be a motivation to raise it higher, so merging!

PR #85, opened 2017-08-27
add lb120 sysinfo to test suite, thanks to githubbuh

rytilahti:


kirichkov (approved):
Ready to merge. Let's ignore the hound for this one!

rytilahti:
Indeed, I think there is a better ways to store those sysinfos (just json files per device?), but this shall do for now.

PR #89, opened 2017-09-14
[CLI] Fix brightness and temperature.

gollo:
Both these command line options raised exceptions either with or without a value. 

Should work as expected now 
```
$ pyhs100 --ip 192.168.0.xxx --bulb brightness 50
Setting brightness to 50
$ pyhs100 --ip 192.168.0.xxx --bulb brightness
Brightness: 50
```


rytilahti:
Thanks for testing and the PR, this shall also go to the release!

PR #90, opened 2017-09-16
add typing hints to make it easier for 3rd party developers to use the library

rytilahti:
On a side note this also changes the behavior of `SmartDevice.features`. Instead of returning None when the feature property is not available, an empty list will be returned to avoid potential crash in `has_emeter`.

The changelog needs to be adjusted accordingly, I hope we can also get this already into 0.3.0 release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: false
 language: python
 python:
+  - "3.3"
   - "3.4"
   - "3.5"
   - "3.6"
--- a/pyHS100/discover.py
+++ b/pyHS100/discover.py
@@ -1,20 +1,24 @@
 import socket
 import logging
 import json
+from typing import Dict
 
-from pyHS100 import TPLinkSmartHomeProtocol, SmartPlug, SmartBulb
+from pyHS100 import TPLinkSmartHomeProtocol, SmartDevice, SmartPlug, SmartBulb
 
 _LOGGER = logging.getLogger(__name__)
 
 
 class Discover:
     @staticmethod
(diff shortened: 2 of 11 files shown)


kirichkov (commented):
This looks very nice and useful, what bothers me is this:

New in version 3.5.

Forgive my lameness but doesn't that mean that the library would require python 3.5 and above?

If that's the case - I think it's not a great idea to implement this. There's a bunch of LTS Linux distros that ship with python 3.3 and 3.4 - Ubuntu 14.04 is my personal use case, which comes with python 3.4.

rytilahti:
It's not a lame question at all! I looked it up, and although PEP 484 is only there since Python 3.5, the used annotation format has been available since Python 3.0 (PEP 3107). 

For some types it is necessary to import them from `typing`module (such as `Optional` and `Dict`). The backported, official(?) version of typing is also available in pip (https://pypi.python.org/pypi/typing) and supports both python 3.3 and 3.4 (Travis has run our tests just fine even on 3.4 :-) I think we need to add that to our dependencies in `setup.py` (currently only in tox.ini for tests).

rytilahti:
The only thing breaking python 3.3 was import of `enum` for the unused `DeviceType` which I removed. Enums are also available through a backported https://pypi.python.org/pypi/enum34 but for the moment it is ok like this. Python 3.2 is EOL'd so I don't see that we should support it at all. Now the library is compatible with all currently supported Python 3 versions (3.3, 3.4, 3.5 and 3.6).

kirichkov:
That's great! Merge it when you feel it's ready!

PR #93, opened 2017-10-06
move has_emeter implementation from SmartDevice to SmartPlug, avoid using features() internally

rytilahti:
Consider `has_emeter` to be a part of the public API also in the future, the implementation specifics are moved into SmartPlug (as that of SmartBulb was already so). This is the last change I would like to get in before releasing 0.3.0.


rytilahti:
As soon as I get an okay for this, I'll merge and make the release 0.3.0! :-)

kirichkov (approved):
OK!

Merge and release!

rytilahti:
Run the tests also on a real device and spotted a couple of issues in tests, so fixed those too.

PR #95, opened 2017-10-06
move SmartDeviceException to SmartDevice, and remove types.py complet…

### LINKED ISSUE #94: types.py clashes with Python standard lib 'types'
Please consider renaming the types.py to device_types.py or similar, as it causes an import error in the click package when running cli.py:

       ImportError: cannot import name 'MappingProxyType'

Reason: Python searches for 'MappingProxyType' in types.py instead of the Python standard library 'types'.

rytilahti:
…ely. fixes #94

### DIFF
--- a/pyHS100/__init__.py
+++ b/pyHS100/__init__.py
@@ -13,9 +13,8 @@ Module-specific errors are raised as `SmartDeviceException` and are expected
 to be handled by the user of the library.
 """
 # flake8: noqa
-from .smartdevice import SmartDevice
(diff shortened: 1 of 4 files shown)


kirichkov (approved):
Can you test it on your device, I'm away from my device, and prefer not to mess with it remotely. Such testing  never seems to end well!

Other then that I see no issues.

rytilahti:
Run the tests also on a real device, thanks for the review!

PR #98, opened 2017-10-11
Added transition period for SmartBulbs

Jakpok:
Changed HSV function for SmartBulbs - now user can set length of transition period (instant or very smooth change is now possible). JSON value "transition_period" is correctly interpreted by TP-Link bulbs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ Turning led to True
 
 ## Bulb-specific commands
 
-At the moment setting brightness, color temperature and color (in HSV) is supported.
+At the moment setting brightness, color temperature and color (in HSV) is supported. You can also set transition period for instant or smooth fades
 The commands are straightforward, so feel free to check `--help` for instructions how to use them.
 
 **Feel free to submit patches as pull requests to add more functionality (e.g. scenes)!**
@@ -204,10 +204,10 @@ if bulb.is_variable_color_temp:
 
 ### Setting the color
 
-Hue is given in degrees (0-360) and saturation and value in percentage.
+Hue is given in degrees (0-360) and saturation and value in percentage. You can also add optional, fourth parameter - transition period (in milliseconds)
 
 ```python
 print(bulb.hsv)
 if bulb.is_color:
-   bulb.hsv = (180, 100, 100) # set to cyan
+   bulb.hsv = (180, 100, 100, 1000) # set to cyan, fade 1 second
 ```
--- a/pyHS100/cli.py
+++ b/pyHS100/cli.py
@@ -16,7 +16,6 @@ from pyHS100 import (SmartDevice,
 
 pass_dev = click.make_pass_decorator(SmartDevice)
 
-
 @click.group(invoke_without_command=True)
 @click.option('--ip', envvar="PYHS100_IP", required=False)
 @click.option('--debug/--normal', default=False)
@@ -165,14 +164,15 @@ def temperature(dev, temperature):
 @click.argument("h", type=click.IntRange(0, 360), default=None)
 @click.argument("s", type=click.IntRange(0, 100), default=None)
 @click.argument("v", type=click.IntRange(0, 100), default=None)
+@click.argument("period", type=click.IntRange(0, 10000), default=500)
 @pass_dev
-def hsv(dev, h, s, v):
+def hsv(dev, h, s, v, period):
     """Get or set color in HSV. (Bulb only)"""
     if h is None or s is None or v is None:
         click.echo("Current HSV: %s" % dev.hsv)
     else:
         click.echo("Setting HSV: %s %s %s" % (h, s, v))
-        dev.hsv = (h, s, v)
+        dev.hsv = (h, s, v, period)
 
 
 @cli.command()
--- a/pyHS100/smartbulb.py
+++ b/pyHS100/smartbulb.py
@@ -112,21 +112,25 @@ class SmartBulb(SmartDevice):
         return hue, saturation, value
 
     @hsv.setter
-    def hsv(self, state: Tuple[int, int, int]):
+    def hsv(self, state: Tuple[int, int, int, int]):
         """
         Sets new HSV, if supported
 
-        :param tuple state: hue, saturation and value (degrees, %, %)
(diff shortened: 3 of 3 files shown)


pyHS100/smartbulb.py
```diff
@@ -112,21 +112,25 @@ def hsv(self) -> Optional[Tuple[int, int, int]]:
...
         """
         Sets new HSV, if supported
 
-        :param tuple state: hue, saturation and value (degrees, %, %)
+        :param tuple state: hue, saturation, value and period (degrees, %, %, milliseconds)
         """
         if not self.is_color:
             return None
-
+        if(len(state)<=3):
```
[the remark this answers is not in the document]
kirichkov: Is the 500ms the bulb's default. I.e. If you don't send any transition period, would it change instantly or indeed happen in 500ms by default?

pyHS100/cli.py
```diff
@@ -165,14 +164,15 @@ def temperature(dev, temperature):
 @click.argument("h", type=click.IntRange(0, 360), default=None)
 @click.argument("s", type=click.IntRange(0, 100), default=None)
 @click.argument("v", type=click.IntRange(0, 100), default=None)
+@click.argument("period", type=click.IntRange(0, 10000), default=500)
```
rytilahti, line 167: Is 10000 a limit enforced by the bulb? If yes, maybe it should also be enforced on SmartBulb class too?

README.md
```diff
@@ -204,10 +204,10 @@ if bulb.is_variable_color_temp:
...
-Hue is given in degrees (0-360) and saturation and value in percentage.
+Hue is given in degrees (0-360) and saturation and value in percentage. You can also add optional, fourth parameter - transition period (in milliseconds)
 
 ```python
 print(bulb.hsv)
 if bulb.is_color:
-   bulb.hsv = (180, 100, 100) # set to cyan
-```
\ No newline at end of file
+   bulb.hsv = (180, 100, 100, 1000) # set to cyan, fade 1 second
```
rytilahti, line 212: I'm starting to dislike the property setter API, as this will make setting it ambiguous (as hsv stands for hue, saturation and value, and not for hue, saturation, value and transition). Would we be better off with providing `set_hsv(h, s, v, transition=500)` and trying to make people to convert to it?

pyHS100/smartbulb.py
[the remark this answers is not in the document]
rytilahti: See my previous commentary on overloading the hsv.

rytilahti (commented):
Thanks for the patch, I left some questions we have to discuss about regarding to the API before merging this in.

README.md
kirichkov: My point was to not set a new default, different than the bulb's inbuilt one. I do agree that overloading it is much better, and unless transition is specified to let the bulb handle the transition period. Maybe even create two setters - set_hsv() and set_hsvp() or set_hsvt() (p is for period, t is for transition, I like the latter better).

pyHS100/smartbulb.py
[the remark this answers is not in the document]
Jakpok: Without transition-period value in json the bulb will make change color smoothly in about 500ms (this is hardware-defined)

pyHS100/cli.py
Jakpok: It seems like the bulb doesn't have any limit. I can remove this range from CLI

README.md
rytilahti: We could just add `set_hsv()` which does the heavy lifting, and provide the property as an API to be used if one does not want to control the transition period. I don't see the value of adding an extra letter to the name.

Would that sound good to you? So something like this:

```
def set_hsv(hue, sat, val, transition=None):
    # do what is currently done in property setter here
    # apply transition if given etc.
```

and 
```
@hsv.setter
def hsv(hsv):
    self.set_hsv(hsv[0], hsv[1], hsv[2])
```
or similar?

pyHS100/cli.py
rytilahti: There must be some kind of limitation, but ok!

README.md
kirichkov: This is great 👍

rytilahti: So, I think we should also then add setters for the rest of the functionality too, in case someone wants to use the non-property versions of them, right? That can be done in a separate PR though. @Jakpok, do you mind to update your PR accordingly?

Jakpok: @rytilahti Ok, I can update my request, I will able to do it during the weekend

rytilahti:
Any news on this?

bacumings:
Hi all - wondering if this request is still active? I noticed this functionality was missing and came across this thread while googling the issue.

tempse:
I tested these changes with an LB130 bulb and can confirm that they work as I expect them to.

Example: Switching from red to greenish light in 10 seconds via the CLI:

```
% pyhs100 --ip <IP> --bulb state
== SmartBulb - LB130(EU) ==
Device state: ON
IP address: <IP>
Brightness: 100
HSV: (0, 100, 255)
Is dimmable: True
Color temperature: 0

# invoking the transition
% pyhs100 --ip <IP> --bulb hsv 140 100 100 10000

% pyhs100 --ip <IP> --bulb state
== SmartBulb - LB130(EU) ==
Device state: ON
IP address: <IP>
HSV: (140, 100, 99)
Is dimmable: True
Brightness: 39
Color temperature: 0
```

I tried a couple of different transition times between 0 and 10000 ms, which seems to work on the tested bulb. Entering a non-integer value results in an error which is accompanied with the hint `Usage: pyhs100 hsv [OPTIONS] H S V [PERIOD]` (nice!).

This PR seems like a useful addition to the package. (For example, in Home Assistant, the [light component](https://www.home-assistant.io/components/light/) implements the `transition` attribute, which has no effect whatsoever on a TP-Link bulb. Propagating the changes over to Home Assistant and making this attribute accessible from there would be great too, in my opinion.)

rytilahti:
Now that the API cleanup is done, it should be much easier to implement this properly. Mind updating the PR?

rytilahti:

Issue #97, opened 2017-10-07
Increase the test coverage


rytilahti:
Increasing the test coverage would be a nice set of tasks for Hacktoberfest if someone wants to pursue that. See https://coveralls.io/github/GadgetReactor/pyHS100?branch=master for possible places for improvement.

peterjgrainger:
Looks like cli.py is at 0%, I'll have a go at that one if you want.  I'll let you know any progress tomorrow :)

rytilahti:
@TheSmokingGnu I'm not sure if it is worth to test the client itself, the library parts are much more important (and probably easier to test?) in my opinion :-) However, if you like to add tests for the cli, that would also be appreciated!

peterjgrainger:
Hrmmm...  Maybe I'll start with smartbulb.py first and then give the cli a go.

rytilahti:
I stumbled upon this http://click.pocoo.org/5/testing/ so maybe testing the cli is not so cumbersome as I thought it would be.

peterjgrainger:
Thanks.  I'm going through the bulb test, pretty straight forward.  I'll
put in a or for that first :)

rytilahti:
This is too generic and the coverage got already better with the referenced PR, so I'm closing this.

PR #99, opened 2017-10-13
created a docker file to aid dev setup

peterjgrainger:
Not sure if this will help but while I was working on the tests for #97 I was having issues setting up so the easiest way I found was to create a docker container so all the runs were consistent.

Hopefully someone else will find it useful

I'll keep working on the tests :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,50 @@
+####################################################
+# Use python 2.x and the latest version of alpine  #
+####################################################
+FROM python:3-alpine
+
+LABEL maintainer=peter@grainger.xyz
+
+###################################################
+# Add a compiler                                  #
+###################################################
+RUN apk update && \
+    apk add --no-cache gcc \
+                       libffi-dev \
+                       openssl-dev \
+                       libxslt-dev \
+                       libxml2-dev \
+                       musl-dev \
+                       linux-headers
+
+###################################################
+# Create somewhere to put the files               #
+###################################################
+RUN mkdir -p /opt/pyHS100
+WORKDIR /opt/pyHS100
+
+###################################################
+# Requirements file first to help cache           #
+###################################################
+COPY requirements.txt .
+RUN pip install -r requirements.txt
+
+###################################################
+# Install dev dependancies                         #
+###################################################
+RUN pip install pytest pytest-cov voluptuous typing
+
+###################################################
+# Copy over the rest.                             #
+###################################################
+COPY ./ ./
+
+###################################################
+# Install everything to the path                  #
+###################################################
+RUN python setup.py install
+
+###################################################
+# Run tests                                         #
+###################################################
+CMD pytest
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,12 @@
 ####################################################
-# Use python 2.x and the latest version of alpine  #
+# Use python 3.x and the latest version of alpine  #
 ####################################################
 FROM python:3-alpine
 
 LABEL maintainer=peter@grainger.xyz
 
 ###################################################
-# Add a compiler                                  #
+# Add all system packages relied upon by python   #
 ###################################################
 RUN apk update && \
     apk add --no-cache gcc \
@@ -30,7 +30,7 @@ COPY requirements.txt .
 RUN pip install -r requirements.txt
 
 ###################################################
-# Install dev dependancies                         #
+# Install dev dependancies                        #
 ###################################################
 RUN pip install pytest pytest-cov voluptuous typing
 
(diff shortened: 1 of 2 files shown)


kirichkov:
This indeed can be useful with Linux testing. However, it won't aid with OS X and Windows testing. I'll let @rytilahti decide whether we should pull it. I don't see any harm in having it in the repo.

peterjgrainger:
Docker for Mac/ windows is pretty good.  I'm using a Mac now and it works
great!  I updated the reader with instructions on installation on all
platforms

peterjgrainger:
I mean readme!

kirichkov:
Yes, but underneath Docker for Mac is running a minimal Linux distro. So technically speaking you are still testing under Linux, although it's a Linux box running under a VM on a Mac ;-)

peterjgrainger:
Yeah, hypervisor.  Surprising it works pretty much the same on all platforms, I’m always amazed :)


On 13 Oct 2017, at 09:02, Georgi Kirichkov <notifications@github.com> wrote:

Yes, but underneath Docker for Mac is running a minimal Linux distro. So technically speaking you are still testing under Linux, although it's a Linux box running under a VM on a Mac ;-)

README.md
```diff
@@ -210,4 +210,20 @@ Hue is given in degrees (0-360) and saturation and value in percentage.
...
+Install Docker for your system [Docker for Mac](https://docs.docker.com/docker-for-mac/), [Docker for Windows](https://docs.docker.com/docker-for-windows/), [Docker for Linux](https://docs.docker.com/engine/installation/#server)
+
+#### Build and run
+
+This command will set up the environment and run the tests on demand.
+
+replace `/absolute/path/to/root/of/directory` after the -v flag below with the absolute path to the directory this project is stored in e.g. `/home/bob.smith/pyHS100`
+
+```shell
+docker build . -t pyhs100 && docker run -v /absolute/path/to/root/of/directory/pyHS100/tests:/opt/pyHS100/pyHS100/tests  pyhs100 pytest
```
rytilahti: Considering that the build is being run inside the main directory, `/absolute/path/to/root/of/directory` part could be replaced with `$(PWD)/pyHS100/tests` to make it easier to use.

Dockerfile
```diff
@@ -0,0 +1,50 @@
+####################################################
+# Use python 2.x and the latest version of alpine  #
```
rytilahti: We do not support python2 anymore :-)

Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+
+###################################################
+# Install everything to the path                  #
+###################################################
+RUN python setup.py install
+
+###################################################
+# Run tests                                         #
+###################################################
+CMD pytest
```
rytilahti: Missing a newline at the end, and those "boxes" are also not properly intended.

README.md
```diff
@@ -210,4 +210,20 @@ Hue is given in degrees (0-360) and saturation and value in percentage.
 print(bulb.hsv)
 if bulb.is_color:
    bulb.hsv = (180, 100, 100) # set to cyan
+```
+
+## Development Setup
+
+### Docker
+
+Install Docker for your system [Docker for Mac](https://docs.docker.com/docker-for-mac/), [Docker for Windows](https://docs.docker.com/docker-for-windows/), [Docker for Linux](https://docs.docker.com/engine/installation/#server)
```
rytilahti: Is this part (especially the links) necessary? I would just move the next subchapter here, and mention in the beginning that the system must have a working docker installation.

rytilahti (changes requested):
Looks good, I added a couple of improvement suggestions. I see also no harm for having this in the repository for those who want to use it, even when I personally prefer to use tox&virtualenvs.

README.md
peterjgrainger: yeah, I had that at first but then realised it won't work on Windows.

I could give one setup for Windows and one for Mac/Linux.

or I could just ignore Windows users :)

Let me know and I'll do that

rytilahti: I would say that if someone is really developing something on Windows they pretty much know what they are doing, considering how difficult it all feels (to me at least). So it should be fine.

kirichkov: I totally agree with @rytilahti 👍 I'm also in the same boat, I think python (and scripting languages in general) can be developed on Windows, but there's so many tools missing it requires a significant upfront investment to bring your environment en par with a UNIX one.

So yeah, if you're developing Python code on Windows, there's probably a good reason for it, and you should already have a good idea how to find your way around.

peterjgrainger:
believe all comments are fixed now.  Let me know if you need anything else changing

rytilahti (approved):

rytilahti:
Looks good, thanks!

PR #100, opened 2017-10-15
Adds a Singleton class

kdschlosser:
This update adds a singleton class so that only a single instance of a device can be created. the benifit to this would be if you connect to a device in on place in your code and you connect to the same device in a different place it does not create a new device instance, instead it returns the already created instance. It makes it easier for doing things like checking if new devices were added to the system. But now also offers the ability to cache specific data sets from the device and know that all stored instances of the device will contain the new cached data set.

I have not tested this code as I do not have a device yet. but I am hoping it should work properly. But i thought it might either spark an idea or maybe be useful

### DIFF
--- a/pyHS100/singleton.py
+++ b/pyHS100/singleton.py
@@ -1,0 +1,33 @@
+
+class ArgsSingleton(type):
+    """
+    A metaclass which permits only a single instance of each derived class
+    sharing the same _class_name for any given set of positional 
+    arguments.
+
+    Attempts to instantiate a second instance of a derived class, or another
+    class with the same _class_name, with the same args will return the
+    existing instance.
+    """
+    _instances = {}
+
+    def __call__(cls, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


kirichkov:
Yikes, lots of tests have failed. Was this so before your changes or is this caused by some changes you introduced?

gollo:
My thinking is that this would be best left to the user of the API to implement as the use case requires, rather than embedded in the API...

rytilahti:
I have to agree with @gollo here, introducing singleton into a library sounds like something to avoid. See also https://stackoverflow.com/a/12755616 . I'm not completely against having some sort of `DeviceManager` class for managing all the devices (based on a config file?) if wanted, but that should be completely separate from the main API and should not be forced on users of the library.

rytilahti:
Closing for inactivity, I do feel that it's not a good design for a library, and if really wanted, should be implemented on the side of the user of the lib.

PR #104, opened 2018-01-06
Add support for host names in addition to IP addresses

khenriks:
It can be handy to be able to specify host names in addition to just IP addresses. This change makes that possible. IP addresses still work, and the functionality is not otherwise changed.

### DIFF
--- a/pyHS100/cli.py
+++ b/pyHS100/cli.py
@@ -18,12 +18,17 @@ pass_dev = click.make_pass_decorator(SmartDevice)
 
 
 @click.group(invoke_without_command=True)
-@click.option('--ip', envvar="PYHS100_IP", required=False)
+@click.option('--ip', envvar="PYHS100_IP", required=False,
+              help='The IP address of the device to connect to. This option '
+              'is deprecated and will be removed in the future; use --host '
+              'instead.')
+@click.option('--host', envvar="PYHS100_HOST", required=False,
+              help='The host name or IP address of the device to connect to.')
 @click.option('--debug/--normal', default=False)
 @click.option('--bulb', default=False, is_flag=True)
 @click.option('--plug', default=False, is_flag=True)
 @click.pass_context
-def cli(ctx, ip, debug, bulb, plug):
+def cli(ctx, ip, host, debug, bulb, plug):
     """A cli tool for controlling TP-Link smart home plugs."""
     if debug:
         logging.basicConfig(level=logging.DEBUG)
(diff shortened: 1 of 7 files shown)


kirichkov:
Thanks for the addition!

We had a similar, but not as comprehensive pull request some time ago (https://github.com/GadgetReactor/pyHS100/pull/32) , and one of the issues I raised there was IPv6 compatibility. Have you checked whether your code works in IPv6 environment?

khenriks:
I've checked, and the answer is yes and no. The code itself works fine with IPv6, but unfortunately the devices themselves don't seem to support IPv6. They do initialize themselves with IPv6 addresses, but they don't listen on port 9999, and in fact ignore (drop) all traffic to that port. If the local DNS environment is set up to return AAAA records for the devices, this means a delay in sending commands while it attempts an IPv6 connection, times out, and reverts to an IPv4 connection. I've reconfigured my own local environment not to provide AAAA records for the plugs.

If it would be better, I can update this pull request to prefer IPv4 instead so that kind of local configuration isn't necessary. It's too bad the devices themselves don't support IPv6 though.

kirichkov:
That's an interesting find. I'd like to get some feedback from @rytilahti on what action to take. I personally have disabled IPv6 as it causes me only problems, without no benefit for now.

rytilahti:
It looks good to me, and I applaud the clean ups for discovery! Could you please split this PR to two parts: 1) host name support and 2) clean ups for easier review?

khenriks:
Sure, I'm happy to do that. The discovery cleanup is already in a separate commit. Are you asking to just have it split into a separate PR?

rytilahti:
Yup, splitting sounds good to me!

khenriks:
Okay, I've created a new pull request for the discovery change only. Once that's closed, I can update this PR.

khenriks:
Alright, I've updated this PR now that the discovery change has been merged.

pyHS100/discover.py
```diff
@@ -61,26 +61,26 @@ def discover(protocol: TPLinkSmartHomeProtocol = None,
...
 
     @staticmethod
-    def discover_single(ip_address: str,
+    def discover_single(host: str,
                         protocol: TPLinkSmartHomeProtocol = None
                         ) -> SmartDevice:
         """
         Similar to discover(), except only return device object for single
-        passed device given by IP address.
+        passed device given by host name.
```
rytilahti: As IP addresses are still accepted, how about "Except only return device object for a single host"?

pyHS100/cli.py
```diff
@@ -19,11 +19,12 @@
 
 @click.group(invoke_without_command=True)
 @click.option('--ip', envvar="PYHS100_IP", required=False)
+@click.option('--host', envvar="PYHS100_HOST", required=False)
```
rytilahti: could you add help texts for these options, pointing out that using IP is deprecated and will likely be removed at some point?

rytilahti (changes requested):
Looks good! I added a couple of small things that should be fixed before the merge.

pyHS100/discover.py
jhalbrecht: >As IP addresses are still accepted, how about "Except only return device object for a single host"?

Could someone elaborate on this? I use pyHS100 on a nat'ed LAN without using a name server. I specify ip address. I've never used a DNS name.

rytilahti: It doesn't change anything in that case, the main point of this PR is to use socket's `create_connection` to allow also passing hostnames instead of IP addresses. I just like to see that the original way of things remains intact (like @khenriks did by not removing `PYHS100_IP` and `--ip`, merely adding a host option besides them, which is great!).

pyHS100/cli.py
khenriks: Okay, sounds good, done.

pyHS100/discover.py
khenriks: Alright, that's done.

@jhalbrecht, just to clarify, this will still work with IP addresses, but it will also work with host names for people who want to use those.

rytilahti:
Looks good, thanks!

rytilahti (approved):

rytilahti:
Looks good, thanks! :+1:

PR #105, opened 2018-01-10
Add missing "future" requirement.

mrstegeman:


rytilahti:
Can you please tell where future is being used? I see no imports from that module in master :-)

mrstegeman:
Hmm... you're right. If you download the 0.3.0 wheel from pypi, it contains a `pyHS100.py` file that requires `future`, but the source package on pypi does not include that file.

https://pypi.python.org/pypi/pyHS100/0.3.0

rytilahti:
Ah, hmm, that's a bit weird. There has been no support for python2 (and thus no need for future) since april last year (and 0.3.0 should have no pyHS100.py at all either). Unfortunately I don't have currently time to investigate what is wrong in there at the moment.

mrstegeman:
I think just uploading a new wheel without that file (maybe making an 0.3.1 release) would do the trick.

PR #106, opened 2018-01-14
Use direct device type discovery for devices

khenriks:
This is more efficient than enumerating all devices and checking the IP.


rytilahti:
Looks and works nicely, thanks a lot!

PR #115, opened 2018-04-08
Update smartplug.py to support dimming in HS220

JsChiSurf:
Switch functions essentially as a "plug" with the addition of supporting dimming,  which can be test for by verifying existence of the 'brightness' array value.

### DIFF
--- a/pyHS100/smartplug.py
+++ b/pyHS100/smartplug.py
@@ -110,18 +110,18 @@ class SmartPlug(SmartDevice):
         elif value > 0 and value <= 100:
             self.turn_on()
             self._query_helper("smartlife.iot.dimmer", "set_brightness",
-                              {"brightness": value})
-        else:	
+                               {"brightness": value})
+        else:
             raise ValueError("Brightness value %s is not valid.", value)
-    
+
     @property
     def is_dimmable(self):
         """
         Whether the switch supports brightness changes
 
-        :return: True if the switch supports brightness changes, False otherwise
+        :return: True if switch supports brightness changes, False otherwise
         :rtype: bool
-        
+
         """
         dimmable = False
         if "brightness" in self.sys_info:
--- a/pyHS100/smartplug.py
+++ b/pyHS100/smartplug.py
@@ -111,8 +111,8 @@ class SmartPlug(SmartDevice):
             return None
 
         if not isinstance(value, int):
-            raise ValueError("Brightness must be integer, 
-                             not of %s.", type(value))
+            raise ValueError("Brightness must be integer, " 
+                             "not of %s.", type(value))
         elif value > 0 and value <= 100:
             self.turn_on()
             self._query_helper("smartlife.iot.dimmer", "set_brightness",
--- a/pyHS100/smartplug.py
+++ b/pyHS100/smartplug.py
@@ -85,8 +85,8 @@ class SmartPlug(SmartDevice):
         """
         Current brightness of the device, if supported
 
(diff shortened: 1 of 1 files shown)


rytilahti:
Without having looked into this that much, I think that the dimmer is more like a bulb than it is a switch, and should therefore belong to smartbulb. Would that make any sense?

JsChiSurf:
It's tough since it's a hybrid. But, it seems to share more in common with the plug library than the bulb, with the only difference being the brightness setting. I have multiple HS200s (non-dimmable switch), and I wouldn't think to use the bulb library for them.

rytilahti:
Ok, sounds fair enough! I'm wondering how shall we do autodetection for these devices (maybe just checking if the brightness is exposed?). I'll add a couple of comments to the code now already though.

pyHS100/smartplug.py
```diff
@@ -80,6 +80,54 @@ def state(self, value: str):
         else:
             raise ValueError("State %s is not valid.", value)
 
+    @property
+    def brightness(self) -> Optional[int]:
+        """
+        Current brightness of the device, if supported
```
rytilahti: Please add the range of returned values here.

pyHS100/smartplug.py
```diff
@@ -80,6 +80,54 @@ def state(self, value: str):
...
+        if not self.is_dimmable:
+            return None
+
+        return int(self.sys_info['brightness'])
+        
+    @brightness.setter
+    def brightness(self, value: int):
+        """
+        Set the new switch brightness level
+
```
rytilahti: Maybe it is worth mentioning here, that this will turn the light on if it isn't already on?

pyHS100/smartplug.py
```diff
@@ -80,6 +80,54 @@ def state(self, value: str):
...
+        :param value: integer between 1 and 100
+        
+        """
+        if not self.is_dimmable:
+            return None
+
+        if not isinstance(value, int):
+            raise ValueError("Brightness must be integer, not of %s.", type(value))
+        elif value > 0 and value <= 100:
```
rytilahti: Just to be sure, the values are between 1-100?

rytilahti (commented):
Looks generally good to me, needs updating the auto discovery & README.md to include the new type, but otherwise this can be merged in my opinion.

pyHS100/smartplug.py
JsChiSurf: Yes, this is correct, I specified valid values to be between 1 and 100.  I did not feel comfortable "assuming" that setting brightness to 0 would turn the light off, versus an explicit call by sending to the "off" command.  I can test this tonight by sending a value of 0 for brightness to see if the "relay_state" automatically switches to 0 when / if brightness is set to 0 before, and if so, allow 0 in the brightness range before this is merged.

I will also update the auto discovery & README.md files accordingly.

JsChiSurf: Okay, I VPN'd home from work and can indeed confirm that allowing a brightness of 0 will NOT turn the switch off and in addition will NOT change whatever the current brightness level is (i.e. if brightness is set to 60, and a value of 0 is passed in, it will stay at 60).  Furthermore, the way things are currently coded, by allowing 0 as a brightness value, this would turn the light on to whatever its previous brightness value is set to, since 0 has no impact.  So, I think things are good they way they are.

rytilahti:
Looks good, let's get this merged. Thanks a lot!

PR #118, opened 2018-04-15
Added option to control devices by device name

kwazel:
Since I have no functional DNS server in my home network and I refuse to remember IP-addresses, I found the possibility to reach devices using their device name (as set using the smartphone App) convenient

### DIFF
--- a/pyHS100/cli.py
+++ b/pyHS100/cli.py
@@ -24,11 +24,14 @@ pass_dev = click.make_pass_decorator(SmartDevice)
               'instead.')
 @click.option('--host', envvar="PYHS100_HOST", required=False,
               help='The host name or IP address of the device to connect to.')
+@click.option('--devicename', envvar="PYHS100_NAME", required=False,
+              help='The device name of the device to connect to.')
+
 @click.option('--debug/--normal', default=False)
 @click.option('--bulb', default=False, is_flag=True)
 @click.option('--plug', default=False, is_flag=True)
 @click.pass_context
-def cli(ctx, ip, host, debug, bulb, plug):
+def cli(ctx, ip, host, devicename, debug, bulb, plug):
     """A cli tool for controlling TP-Link smart home plugs."""
     if debug:
         logging.basicConfig(level=logging.DEBUG)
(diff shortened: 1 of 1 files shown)


pyHS100/cli.py
```diff
@@ -80,6 +89,25 @@ def discover(ctx, timeout, discover_only):
...
+    click.echo("Trying to discover %s using %s attempts of %s seconds" %
+               (devicename, attempts, timeout))
+    for attempt in range(1, attempts):
+        print("Attempt %s of %s" % (attempt, attempts))
+        found_devs = Discover.discover(timeout=timeout).items()
+        for ip, dev in found_devs:
+            if dev.alias.lower() == devicename.lower():
+                host = dev.host
+                return host
+    return None
```
rytilahti: Should this rather raise a corresponding exception instead of returning none?

pyHS100/cli.py
```diff
@@ -80,6 +89,25 @@ def discover(ctx, timeout, discover_only):
...
+@click.option('--timeout', default=3, required=False)
+@click.option('--attempts', default=10, required=False)
+@click.option('--devicename', required=True)
+@click.pass_context
+def find_host_from_device_name(ctx, timeout, attempts, devicename):
+    """Discover devices in the network."""
+    host = None
+    click.echo("Trying to discover %s using %s attempts of %s seconds" %
+               (devicename, attempts, timeout))
+    for attempt in range(1, attempts):
```
rytilahti, line 102: Is there really a need to do multiple attempts? When does that happen? 10 attempts with timeout of 3 seconds sounds a bit much, doesn't it?

pyHS100/cli.py
```diff
@@ -80,6 +89,25 @@ def discover(ctx, timeout, discover_only):
     return found_devs
 
 
+@cli.command()
+@click.option('--timeout', default=3, required=False)
+@click.option('--attempts', default=10, required=False)
+@click.option('--devicename', required=True)
+@click.pass_context
+def find_host_from_device_name(ctx, timeout, attempts, devicename):
```
rytilahti, line 97: Does this have to be a command? Is there some usecase where someone would call this from console directly?

rytilahti (commented):
Although I would probably prefer a configuration file by myself (that would avoid doing the discovery in those cases where the IP address stays the same), I can see how this can be useful and would be fine merging this after the comments are handled.

pyHS100/cli.py
kirichkov: I agree with @rytilahti. I think a default timeout of 1 second and 3 attempts sounds acceptable to me.

pyHS100/cli.py
kirichkov: I see the logic of returning `None`. If a device can't be found via name, the cli will continue searching for a device using discovery. On the other hand unsuccessful discovery also doesn't raise an exception, so I think we should adhere here to `discover`'s behavior.

PR #119, opened 2018-04-25
Ability to Switch Bulb on to Specific Brightness/Color/Temperature

edrose:
This Pull Request fixes an issue I came across where it was impossible to switch the LB130 on to a specific color, temperature or brightness instantly. Instead, the behaviour was for the default state of the light to come on, before changing to the desired brightness after a delay of about a second. This posed a problem in a project I worked on whereby the initial brightness would hurt the eyes of someone used to the dark, before the comfortable dim red light state became active.

It simply adds two lines to the `hsv`, `color_temp` and  `brightness` setters which tells the bulb to switch on if not already and to ignore the default state.

### DIFF
--- a/pyHS100/smartbulb.py
+++ b/pyHS100/smartbulb.py
@@ -122,6 +122,8 @@ class SmartBulb(SmartDevice):
             return None
 
         light_state = {
+            "on_off": 1,
+            "ignore_default": 1,
             "hue": state[0],
             "saturation": state[1],
             "brightness": int(state[2] * 100 / 255),
@@ -157,6 +159,8 @@ class SmartBulb(SmartDevice):
             return None
 
         light_state = {
+            "on_off": 1,
+            "ignore_default": 1,
             "color_temp": temp,
         }
         self.set_light_state(light_state)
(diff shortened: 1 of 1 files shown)


kirichkov:
I think you're missing a commit.

edrose:
@kirichkov Just one commit, adds extra options to the requests for setting HSV, brightness or temperature which is just 6 lines. Is there something I'm missing?

kirichkov:
My bad! "adds two lines to the hsv, color_temp and brightness **setters**" - I presumed there's more code to be added to the setters.

@rytilahti should take a look, since he knows the smart bulb code. Your pull request will definitely change the current behavior and this might be a problem with integrations.

edrose:
Sorry I wasn't clear enough!

Yes it will change the behaviour since currently calls to the affected functions will do nothing if the bulb is off. The changes I made will make those functions work regardless of the state of the bulb, so I guess that could cause a problem depending on how the code is used in other projects.

rytilahti:
I'm not really familiar that code either, but if changing those variables do not change the stored values of the bulb (I suppose one should change the values under `dft_on_state` to make offline changes?), I think this is fine.

If offline changes are wanted, we could think of having a way (parameter to `__init__`?) to do that.

rytilahti:
Ping @edrose, I think the best solution here is to add a new `set_<x>` methods taking a `force` parameter (defaulting to `False`), which will add those two parameters when called with `True`. This will also be helpful for #98, so if you are still interested in this please let us know! :-)

rytilahti:

PR #122, opened 2018-06-17
Add bulb valid temperature range

titilambert:
Linked to https://github.com/home-assistant/home-assistant/pull/15020

### DIFF
--- a/pyHS100/cli.py
+++ b/pyHS100/cli.py
@@ -153,13 +153,15 @@ def brightness(dev, brightness):
 
 
 @cli.command()
-@click.argument("temperature", type=click.IntRange(2700, 6500), default=None,
+@click.argument("temperature", type=click.IntRange(2500, 9000), default=None,
                 required=False)
 @pass_dev
 def temperature(dev, temperature):
     """Get or set color temperature. (Bulb only)"""
     if temperature is None:
         click.echo("Color temperature: %s" % dev.color_temp)
+        if dev.valid_temperature_range != (0, 0):
+            click.echo("(min: %s, max: %s)" % dev.valid_temperature_range)
     else:
         click.echo("Setting color temperature to %s" % temperature)
         dev.color_temp = temperature
--- a/pyHS100/smartbulb.py
+++ b/pyHS100/smartbulb.py
@@ -1,5 +1,11 @@
 from pyHS100 import SmartDevice
+import re
 from typing import Any, Dict, Optional, Tuple
+
+TPLINK_KELVIN = {'LB130': (2500, 9000),
+                 'LB120': (2700, 6500),
+                 'LB230': (2500, 9000),
+                 'KB130': (2500, 9000)}
 
 
 class SmartBulb(SmartDevice):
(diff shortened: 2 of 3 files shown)


pyHS100/smartbulb.py
```diff
@@ -1,6 +1,9 @@
 from pyHS100 import SmartDevice
+import re
 from typing import Any, Dict, Optional, Tuple
 
+TPLINK_KELVIN = {'LB130': (2700, 9000),
+                 'LB120': (2700, 6500)}
```
rytilahti: While you are at it, please add information also from other bulbs: https://www.tp-link.com/us/home-networking/smart-home/smart-bulbs

pyHS100/cli.py
```diff
@@ -160,6 +160,7 @@ def temperature(dev, temperature):
     """Get or set color temperature. (Bulb only)"""
     if temperature is None:
         click.echo("Color temperature: %s" % dev.color_temp)
+        click.echo("(min: %s, max: %s)" % dev.valid_temperature_range)
```
rytilahti: Did you test that this works (considering it's a tuple what gets returned)?

rytilahti: Please also handle the case where `is_variable_color_temp` is False.

pyHS100/smartbulb.py
```diff
@@ -78,6 +81,20 @@ def is_variable_color_temp(self) -> bool:
...
+    @property
+    def valid_temperature_range(self) -> Tuple[int, int]:
+        """
+        Returns the white temperature range (in Kelvin)
+        depending on the bulb model
+
+        :return: White temperature range in Kelvin (minimun, maximum)
+        :rtype: tuple
+        """
+        for model, temp_range in TPLINK_KELVIN.items():
```
rytilahti: I think you should check `is_variable_color_temp` and return (maybe 0, 0) when not available.

rytilahti: Setter for temperature could also be adopted to enforce these limits by clamping the values (and adapting the tests accordingly).

pyHS100/smartbulb.py
titilambert: done !

rytilahti: How about LB110, LB110, LB200, and LB230 (although LBx series is new to me, no idea if it's supported, but it would make sense to have these already in place anyway)?

edit: looks like white only bulbs (LB110, LB100, LB200) do not support color temp, so only LB230 is missing, but they could be set `(0, 0)` for the sake of completeness.

pyHS100/cli.py
titilambert: Yes and done !

pyHS100/smartbulb.py
titilambert: Do you prefer (0,0) or False ?

pyHS100/smartbulb.py
```diff
@@ -78,6 +84,20 @@ def is_variable_color_temp(self) -> bool:
...
+            if re.match(model, self.sys_info['model']):
+                return temp_range
+        return False
```
rytilahti: Please return (0, 0) instead of False (as `-> Tuple[int, int]` waits for a tuple and it's better not to make confusion between types).

titilambert: ok !

titilambert:
I think it's one @rytilahti ! Did I miss something ?

pyHS100/smartbulb.py
```diff
@@ -155,6 +177,10 @@ def color_temp(self, temp: int) -> None:
         if not self.is_variable_color_temp:
             return None
 
+        if temp < self.valid_temperature_range[0] or \
+                temp > self.valid_temperature_range[1]:
+            raise ValueError
```
rytilahti: Please add an error message indicating the that the valid range was exceeded and add a tests for both error cases (too small and too large temperature), then it's good to go! :-)

titilambert: Done !

rytilahti:
Great, thanks for the patch! I'll prepare a new release asap.

PR #124, opened 2018-07-22
Fix bug that changed brightness at each HSV update

tempse:
The HSV setter should accept a percentage for the brightness
value but actually assumed the brightness to be in absolute values
between 1 and 255.
This resulted in brightness reductions at each HSV update, in
steps of 100% -> 100/255=39% -> 39/255=15% -> ... (see also
https://github.com/home-assistant/home-assistant/issues/15582,
where I originally reported this bug).

PR #125, opened 2018-08-03
Support Unicode strings in encrypt/decrypt

amelchio:
I have a problem where my HS110 smartplugs with localized names show up with charset errors in Home Assistant.

This PR fixes that by adding utf-8 encode/decode to the protocol encryption functions.


rytilahti:
Looks good and works fine here, thanks! I'll add a command to set the name to the cli in another PR. I'll create a new release after we get #124 merged.

PR #126, opened 2018-08-08
add 'alias' command for querying and setting the alias

rytilahti:


kirichkov:
Perhaps add a test or exclude the alias sub-command  from coverage testing?

rytilahti:
I excluded all non-library files (including the client and test files) from the coverage.

edit: looks like tox does not support python 3.4 anymore, so I think it's going to be dropped from testing (although it may still working on it, obviously).

```
  File "/home/travis/virtualenv/python3.3.6/bin/tox", line 7, in <module>
    from tox import cmdline
  File "/home/travis/virtualenv/python3.3.6/lib/python3.3/site-packages/tox/__init__.py", line 10, in <module>
    from pkg_resources import DistributionNotFound, get_distribution
  File "/home/travis/virtualenv/python3.3.6/lib/python3.3/site-packages/pkg_resources/__init__.py", line 85, in <module>
    raise RuntimeError("Python 3.4 or later is required")
RuntimeError: Python 3.4 or later is required
```

kirichkov:
Hmmm. I think something else changed. Travis is showing a decreese in coverage in files that are not even touched in this PR.

I'd say go ahead and merge it.

rytilahti:
It is because this contains also a commit to only calculate coverage on the library files (e.g. ignoring the tests and the cli). The way the coverage used to be calculated was thus giving out false coverage percent, so I'm merging this now. Thanks for the review!

PR #127, opened 2018-08-27
Use of alias instead of IP address or hostname

kwazel:
Some people will have a network where TP-link devices like power plugs do not have a static ip-address and do not have the luxury of a hostname. In this case eacht TP-link device has a device name, or alias. This can be used to identify the smart plugs and bulbs uniquely. Using the parameter
--alias
instead of --ip or --hostname, the cli will attempt to discover a device with this alias (case insensitive). Upon finding such a device, it will use this device, e.g. `cli.py --alias kalle --plug off` will switch a smart plug with the alias "Kalle" or "kalle" off.
`cli.py --alias "kalle" --plug on` will switch the device on again. A little bit confusing will be
`cli.py --alias "kalle" alias "kalle anka"` which will change the device name, alias, of a device with the alias "kalle". The new alias will be "kalle anka". When undesired, this can be circumvented by:

- changing `--ip`, `--hostname` or `--alias` into `--byip`, `--byhostname` or `--byalias`
- changing `alias` as parameter into e.g. `device`
- proper documentation

### DIFF
--- a/pyHS100/cli.py
+++ b/pyHS100/cli.py
@@ -48,7 +48,11 @@ def cli(ctx, ip, host, alias, debug, bulb, plug):
         click.echo("Alias is given, using discovery to find host %s" %
                    alias)
         host = find_host_from_alias(alias=alias)
-        click.echo("Found hostname is {}".format(host))
+        if host:
+            click.echo("Found hostname is {}".format(host))
+        else:
+            click.echo("No device with name {} found".format(alias))
+            return
 
     if host is None:
         click.echo("No host name given, trying discovery..")
--- a/pyHS100/cli.py
+++ b/pyHS100/cli.py
@@ -24,14 +24,14 @@ pass_dev = click.make_pass_decorator(SmartDevice)
               'instead.')
 @click.option('--host', envvar="PYHS100_HOST", required=False,
               help='The host name or IP address of the device to connect to.')
-@click.option('--devicename', envvar="PYHS100_NAME", required=False,
-              help='The device name of the device to connect to.')
+@click.option('--alias', envvar="PYHS100_NAME", required=False,
+              help='The device name, or alias, of the device to connect to.')
 
 @click.option('--debug/--normal', default=False)
 @click.option('--bulb', default=False, is_flag=True)
 @click.option('--plug', default=False, is_flag=True)
 @click.pass_context
-def cli(ctx, ip, host, devicename, debug, bulb, plug):
+def cli(ctx, ip, host, alias, debug, bulb, plug):
     """A cli tool for controlling TP-Link smart home plugs."""
     if debug:
         logging.basicConfig(level=logging.DEBUG)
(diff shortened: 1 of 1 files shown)


kwazel:
I have requested this change before, but back then (April) I did not find time to actually implement the proposed changes (I was actually fighting with the unit tests, which I am now able to run but do not cover the cli very well).

pyHS100/cli.py
```diff
@@ -41,11 +44,17 @@ def cli(ctx, ip, host, debug, bulb, plug):
     if ip is not None and host is None:
         host = ip
 
+    if alias is not None and host is None:
+        click.echo("Alias is given, using discovery to find host %s" %
+                   alias)
+        host = find_host_from_alias(alias=alias)
+        print("Found hostname is {}".format(host))
+
+
```
rytilahti: We do not have hound-ci active at the moment (which i hope we'll get enabled soonish) to notify about code-style issues, but here is an extra newline.

pyHS100/cli.py
```diff
@@ -72,14 +81,28 @@ def discover(ctx, timeout, discover_only):
     click.echo("Discovering devices for %s seconds" % timeout)
     found_devs = Discover.discover(timeout=timeout).items()
     if not discover_only:
-        for ip, dev in found_devs:
+        for _, dev in found_devs:
```
rytilahti: I think it's better to leave it as it was, as it makes it easy for a reader to see what it is behind the variable. So either revert this change, or use `found_devs.items()` and drop it altogether.

pyHS100/cli.py
```diff
@@ -72,14 +81,28 @@ def discover(ctx, timeout, discover_only):
...
     return found_devs
 
 
+def find_host_from_alias(alias, timeout=1, attempts=3):
+    """Discover a device identified by its alias"""
+    host = None
+    click.echo("Trying to discover %s using %s attempts of %s seconds" %
+               (alias, attempts, timeout))
+    for attempt in range(1, attempts):
+        print("Attempt %s of %s" % (attempt, attempts))
```
rytilahti: Use `click.echo` instead of `print`.

pyHS100/cli.py
```diff
@@ -72,14 +81,28 @@ def discover(ctx, timeout, discover_only):
...
+        found_devs = Discover.discover(timeout=timeout).items()
+        for _, dev in found_devs:
```
rytilahti: See my above note about anonymous variables.

pyHS100/cli.py
```diff
@@ -72,14 +81,28 @@ def discover(ctx, timeout, discover_only):
...
+            if dev.alias.lower() == alias.lower():
+                host = dev.host
+                return host
+    return None
+
```
rytilahti, line 108: This should have two empty lines (between functions as per pep8).

rytilahti (changes requested):
Looks good to me, thanks for the contribution! I'm unsure if there's a need to have a envvar for this, but that's fine. I added a couple of very small, easy-to-fix issues that should be changed before merging this.

pyHS100/cli.py
kwazel: I see your point, I was just following Pylints advice.

pyHS100/cli.py
```diff
@@ -41,11 +44,16 @@ def cli(ctx, ip, host, debug, bulb, plug):
     if ip is not None and host is None:
         host = ip
 
+    if alias is not None and host is None:
+        click.echo("Alias is given, using discovery to find host %s" %
+                   alias)
+        host = find_host_from_alias(alias=alias)
+        click.echo("Found hostname is {}".format(host))
```
rytilahti: Could you add a check if the returned host is `None` and return if so, I'd expect that when using `--alias` it should bail out if it cannot discover the device instead of continuing. Otherwise this looks good to be merged :-)

rytilahti (approved):

rytilahti:
Great, thanks! :tada:

PR #129, opened 2018-09-07
Add reboot command to restart the device

### LINKED ISSUE #111: "Reboot" command
Why not?

Turn off, check state, when off, turn on maybe even with an optional time to off?

rytilahti:
Add reboot command. Takes delay (defaulting to 1) for how long to wait before doing the reboot. Setting this to 0 causes an instant reboot with no response from the device, causing an exception being thrown.

```
Usage: pyhs100 reboot [OPTIONS]

  Reboot the device.

Options:
  --delay INTEGER
  --help           Show this message and exit.
```

Fixes #111

### DIFF
--- a/pyHS100/cli.py
+++ b/pyHS100/cli.py
@@ -255,5 +255,13 @@ def off(plug):
     plug.turn_off()
 
 
+@cli.command()
+@click.option("--delay", default=1)
+@pass_dev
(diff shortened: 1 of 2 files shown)


kirichkov (approved):
I think it's good to go. Merge at your convenience!

PR #137, opened 2018-10-20
Add support for HS300 power strip

### LINKED ISSUE #131: Support HS300 Power Strip
Would like support for the Power Strip:
https://www.tp-link.com/us/products/details/cat-5516_HS300.html

Discover currently fails:
```
pi@rpi3-2:~/development/pyHS100 $ python3 pyHS100/cli.py 
No host name given, trying discovery..
Discovering devices for 3 seconds
== TP-LINK_Power Strip_2CA9 - HS300(US) ==
OFF
Host/IP: 192.168.86.45
  File "pyHS100/cli.py", line 267, in <module>
    cli()
  File "/home/pi/.local/lib/python3.5/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/home/pi/.local/lib/python3.5/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/pi/.local/lib/python3.5/site-packages/click/core.py", line 1043, in invoke
    return Command.invoke(self, ctx)
  File "/home/pi/.local/lib/python3.5/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/pi/.local/lib/python3.5/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/home/pi/.local/lib/python3.5/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "pyHS100/cli.py", line 59, in cli
    ctx.invoke(discover)
  File "/home/pi/.local/lib/python3.5/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/home/pi/.local/lib/python3.5/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "pyHS100/cli.py", line 89, in discover
    ctx.invoke(state)
  File "/home/pi/.local/lib/python3.5/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/home/pi/.local/lib/python3.5/site-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
  File "/home/pi/.local/lib/python3.5/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/home/pi/.local/lib/python3.5/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "pyHS100/cli.py", line 129, in state
    for k, v in dev.state_information.items():
  File "/usr/local/lib/python3.5/dist-packages/pyHS100/smartplug.py", line 206, in state_information
    'On since': self.on_since
  File "/usr/local/lib/python3.5/dist-packages/pyHS100/smartplug.py", line 200, in on_since
    datetime.timedelta(seconds=self.sys_info["on_time"])
TypeError: unsupported type for timedelta seconds component: NoneType
```

Because on_time is down inside children:
```
== System info ==
defaultdict(<function SmartDevice.sys_info.<locals>.<lambda> at 0x767d35d0>,
            {'alias': 'TP-LINK_Power Strip_2CA9',
             'child_num': 6,
             'children': [{'alias': 'Home Plug 1',
                           'id': 'xx00',
                           'next_action': {'type': -1},
                           'on_time': 63142,
                           'state': 1},
                          {'alias': 'Home Plug 2',
                           'id': 'xx01',
                           'next_action': {'type': -1},
                           'on_time': 0,
                           'state': 0},
                          {'alias': 'Home Plug 3',
                           'id': 'xx02',
                           'next_action': {'type': -1},
                           'on_time': 0,
                           'state': 0},
                          {'alias': 'Home Plug 4',
                           'id': 'xx03',
                           'next_action': {'type': -1},
                           'on_time': 0,
                           'state': 0},
                          {'alias': 'Home Plug 5',
                           'id': 'xx04',
                           'next_action': {'type': -1},
                           'on_time': 0,
                           'state': 0},
                          {'alias': 'Home Plug 6',
                           'id': 'xx05',
                           'next_action': {'type': -1},
                           'on_time': 0,
                           'state': 0}],
             'deviceId': 'xxx',
             'feature': 'TIM:ENE',
             'hwId': 'xxx',
             'hw_ver': '1.0',
             'latitude_i': xxx,
             'led_off': 0,
             'longitude_i': -xxx,
             'mac': 'xxx,
             'mic_type': 'IOT.SMARTPLUGSWITCH',
             'model': 'HS300(US)',
             'oemId': 'xxx',
             'rssi': -45,
             'sw_ver': '1.0.6 Build 180627 Rel.081000',
             'updating': 0})
```

This is the only device I have.  I can fork and work on it, but may need some guidance on how a multi-plug device should be handled.

jimboca:
These changes are from myself and @newAM please let me know what you think before all changes are completed.

Fixes #131


pyHS100/smartdevice.py
```diff
@@ -127,6 +134,17 @@ def _query_helper(self,
 
         return result
 
+    def _index_to_id(self, index: int) -> str:
```
kirichkov: Is this really necessary? Given that the library is developer-oriented, it could be handled in the CLI, and whoever's using it to implement this themselves.

pyHS100/smartplug.py
```diff
@@ -35,7 +35,7 @@ def __init__(self,
                  host: str,
                  protocol: 'TPLinkSmartHomeProtocol' = None) -> None:
         SmartDevice.__init__(self, host, protocol)
-        self.emeter_type = "emeter"
+        self._type = "emeter"
```
kirichkov, line 39: I don't quite get why this is necessary. Perhaps there are other changes, related, that are not part of this pull yet?

kirichkov:
Overall this looks good to me, several more things:
1. Fix the hound's comments
2. You'll need to write tests, to ensure that a future change doesn't break it.
3. I noticed the files' permissions got an "execute" bit set, you'd need to remove that bit.

Keep up the good work! 👍

pyHS100/smartdevice.py
jimboca: Either way is fine but just seemed better to always allow passing the index instead of child id.

pyHS100/smartplug.py
jimboca: I'm not sure either, I think it was added by @newAM

jimboca:
Thanks for taking a look!  Hopefully will have some more time to work on it this weekend.

pyHS100/smartplug.py
newAM: git blame shows you added it

kirichkov: Maybe you deleted “emeter” by mistake when editing. Let’s revert it if it’s not clear why it was changed.

jimboca: thanks guys, no idea how I did that, I'll put it back.

pyHS100/discover.py
```diff
@@ -52,6 +52,7 @@ def discover(protocol: TPLinkSmartHomeProtocol = None,
                 ip, port = addr
                 info = json.loads(protocol.decrypt(data))
                 device_class = Discover._get_device_class(info)
+                print("device_class={}".format(device_class))
```
rytilahti:
```suggestion
```

pyHS100/discover.py
```diff
@@ -96,9 +99,14 @@ def _get_device_class(info: dict) -> Type[SmartDevice]:
             else:
                 _LOGGER.error("Unable to find the device type field!")
                 type = "UNKNOWN"
+            if 'child_num' in sysinfo:
```
rytilahti:
```suggestion
```

pyHS100/discover.py
```diff
@@ -96,9 +99,14 @@ def _get_device_class(info: dict) -> Type[SmartDevice]:
...
+                child_num = sysinfo['child_num']
```
rytilahti:
```suggestion
```

pyHS100/smartdevice.py
```diff
@@ -89,23 +89,30 @@ def __init__(self,
...
-                      arg: Optional[Dict] = None) -> Any:
+                      arg: Optional[Dict] = None,
+                      index: int = None) -> Any:
         """
         Helper returning unwrapped result object and doing error handling.
 
         :param target: Target system {system, time, emeter, ..}
         :param cmd: Command to execute
         :param arg: JSON object passed as parameter to the command
+        :param index:  optional index number of the child ID for context
```
rytilahti: one whitespace too much here. the description is not also very clear, maybe something like "Plug index on multi-plug devices"?

pyHS100/smartdevice.py
```diff
@@ -127,6 +134,17 @@ def _query_helper(self,
...
         return result
 
+    def _index_to_id(self, index: int) -> str:
+        """
+        Returns the child ID for the given plug index
+
+        :param index: plug index (1 based, not zero based)
+        :raises SmartDeviceException: on error
+        :return: child ID string
+        :rtype: datetime
```
rytilahti: Wrong data type. Should the return value be also int instead of a string?

pyHS100/smartstrip.py
```diff
@@ -0,0 +1,256 @@
...
+        Retrive current energy readings from device.
+
+        :returns: current readings or False
+        :rtype: dict, None
+                  None if device has no energy meter or error occured
+        :raises SmartDeviceException: on error
+        """
+        if not self.has_emeter:
+            return None
+        # Should we build a dict with total and children?
```
rytilahti: I think having an interface to get information from all the plugs in a one go is a good idea. See my general commend on the interface.

pyHS100/smartstrip.py
```diff
@@ -0,0 +1,256 @@
...
+        Whether the switch supports brightness changes
+
+        :return: True if switch supports brightness changes, False otherwise
+        :rtype: bool
+
+        """
+        dimmable = False
+        if "brightness" in self.sys_info:
+            dimmable = True
+        return dimmable
```
rytilahti: Same as above, is this a global functionality, or specific for a single socket?

pyHS100/smartstrip.py
```diff
@@ -0,0 +1,256 @@
...
+
+        if not isinstance(value, int):
+            raise ValueError("Brightness must be integer, "
+                             "not of %s.", type(value))
+        elif value > 0 and value <= 100:
+            self.turn_on()
+            self._query_helper("smartlife.iot.dimmer", "set_brightness",
+                               {"brightness": value})
+        else:
+            raise ValueError("Brightness value %s is not valid.", value)
```
rytilahti: Is this for setting the brightness of all the sockets in a one go?

rytilahti (commented):
This looks to be a great start, thanks a lot for the work! I'll add some issues not touched by @kirichkov here.

I added some short commentary inline, but wanted to highlight how I think the multi-plug systems should be handled in general. 
The API based on the code I'm reading here looks very similar to those of plugs, so I think it's best to avoid code duplication and handle single "sockets" as instances of the SmartPlug class.
This will leave `SmartStrip` to serve as a container/helper (and in some cases as a shortcut, see e.g. my comment on emeter) for them.

This would involve the following changes:
1. Removing single-plug-only properties, or better, converting them to report back a dictionary of  `(SmartPlug, result)` (for emeter, led, on since, maybe more?) from `SmartStrip`.
2. Extending SmartPlug as necessary to allow the new functionality. This requires probably some changes to `sysinfo` to return "indexed" values where necessary.
3. Deciding what actions are reasonable to have on the scope of a whole strip. For example, instead of having separate `turn_on_plug`, the `turn_on` of the `SmartStrip` should simply accept a parameter `index` to decide on whether to turn on just a single socket or all of them. This same applies to emeters and leds likely too -- if called with an index, simply returning the information from the contained `SmartPlug` is expected, but if called without an index, information from all the sockets is to be returned in a dict of (`SmartPlug`, `<value>`).

What do you think, does this sound reasonable?

newAM:
1. I would be on board for that.  From what I have seen the only unique child properties are id, state, alias, on_time, and emeter.  LED control is done by the parent for the HS300, I imagine it would be the same for the HS107 then.
2. I think the best general solution would be to add the parent and child ID to the plug class constructor.  If the child ID is set, then it would append the child ID context to the request for state, alias, on_time, and emeter.
3. I like that solution.

@jimboca I can work on this over the next few days, could you give me write access to your branch so that I can merge mine in for this pull request?

jimboca:
> @jimboca I can work on this over the next few days, could you give me write access to your branch so that I can merge mine in for this pull request?

I added you last night. Your invite is still waiting?  https://github.com/jimboca/pyHS100/invitations

jimboca:
@newAM I gave you permission to mine, change away!  Let me know when it's
at a point and I can try to help again when it's there.

jimboca:
> @newAM I gave you permission to mine, change away! Let me know when it's at a point and I can try to help again when it's there.

That's weird, I added that comment last night but it didn't show up until today...

rytilahti:
I removed the duplicates, looks like github has been having some issues (I received the commentary per e-mail though..).

rytilahti:
> 1. I would be on board for that.  From what I have seen the only unique child properties are id, state, alias, on_time, and emeter.  LED control is done by the parent for the HS300, I imagine it would be the same for the HS107 then.

Ok! We can make adjustments as necessary after someone comes and reports back how the HS107 works, but I would expect it to work similarly.

> 2. I think the best general solution would be to add the parent and child ID to the plug class constructor.  If the child ID is set, then it would append the child ID context to the request for state, alias, on_time, and emeter.

Sounds good!

newAM:
Yikes sorry about those duplicates.  The error message indicated the message was not sent so I pushed the button a few times.  Looks like they all went through.

I got the invite @jimboca, thanks!

newAM:
Changes Made:
* Common methods are inherited from SmartPlug, child-specific methods are overridden
  * SmartStrip inherits from SmartPlug
  * SmartStrip creates a SmartPlug for each child plug
* Addressed some of the hounds comments

Let me know what you guys think.  Not entirely confident I am on the right track here.

ToDo:
* tests
* not sure what to do about the SmartStrip.State property
* remove +x on files

pyHS100/discover.py
```diff
@@ -3,7 +3,8 @@
 import json
 from typing import Dict, Type
 
-from pyHS100 import TPLinkSmartHomeProtocol, SmartDevice, SmartPlug, SmartBulb, SmartStrip
+from pyHS100 import TPLinkSmartHomeProtocol, SmartDevice, SmartPlug, SmartBulb
+from pyHS100 import SmartStrip
```
rytilahti: Please append to the existing import, `from pyHS100 import (xxx, yyy)` will allow line breaks.

pyHS100/discover.py
```diff
@@ -87,6 +88,7 @@ def discover_single(host: str,
     @staticmethod
     def _get_device_class(info: dict) -> Type[SmartDevice]:
         """Find SmartDevice subclass for device described by passed data."""
+        print("got info: {}".format(info))
```
rytilahti:
```suggestion
```

pyHS100/smartstrip.py
```diff
@@ -0,0 +1,155 @@
...
+        for plug in range(self.num_children):
+            relay_state = self.sys_info["children"][plug]["state"]
+
+            if relay_state == 0:
+                plug_states.append(SmartStrip.SWITCH_STATE_OFF)
+            if relay_state == 1:
+                plug_states.append(SmartStrip.SWITCH_STATE_ON)
+            else:
+                _LOGGER.warning("Unknown state %s returned.", relay_state)
+                plug_states.append(SmartStrip.SWITCH_STATE_UNKNOWN)
```
rytilahti: I'd prefer that this would be a dict of `{index: state}`. Also, you can check for the state with `if relay_state:`.

pyHS100/smartdevice.py
```diff
@@ -548,6 +556,21 @@ def state_information(self) -> Dict[str, Any]:
...
+
+        :return: number of children
+        :rtype: int
+        :raises SmartDeviceException: on error
+        """
+        child_num = self.sys_info["child_num"]
+        if child_num:
+            return int(child_num)
+        else:
+            return 0
```
rytilahti: When does the else happen?

pyHS100/smartstrip.py
```diff
@@ -0,0 +1,155 @@
...
+    Errors reported by the device are raised as SmartDeviceExceptions,
+    and should be handled by the user of the library.
+
+    Note:
+    The library references the same structure as defined for the D-Link Switch
+    """
+    # switch states
+    SWITCH_STATE_ON = 'ON'
+    SWITCH_STATE_OFF = 'OFF'
+    SWITCH_STATE_UNKNOWN = 'UNKNOWN'
```
rytilahti: These can be reused directly from `SmartPlug` (`SmartPlug.SWITCH_STATE_ON`) and so on.

pyHS100/smartstrip.py
```diff
@@ -0,0 +1,155 @@
...
+    def is_on(self, index: int) -> bool:
+        """
+        Returns whether device is on.
+
+        :param index: plug index
+        :return: True if device is on, False otherwise
+        """
+        return self.plug[index].is_on()
+
+    def turn_on(self, index: int):
```
rytilahti:
```suggestion
    def turn_on(self, *, index: int = 0):
```

pyHS100/smartstrip.py
```diff
@@ -0,0 +1,155 @@
...
+        """
+        Turns an outlet on
+
+        :param index: plug index
+        :raises SmartDeviceException: on error
+        """
+        self.plug[index].turn_on()
```
rytilahti: See my above suggestion. I'd say that we enforce the keyword argument here for the index, or default to turning all the plugs on and off. What do you think?

pyHS100/smartstrip.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def on_since(self, index: int) -> datetime.datetime:
+        """
+        Returns pretty-printed on-time
+
+        :param index: plug index
+        :return: datetime for on since
+        :rtype: datetime
+        """
+        return self.plug[index].on_since
```
rytilahti: Same as above, returning everything if index is not specified.

pyHS100/smartstrip.py
```diff
@@ -0,0 +1,155 @@
...
+            if relay_state == 1:
+                plug_states.append(SmartStrip.SWITCH_STATE_ON)
+            else:
+                _LOGGER.warning("Unknown state %s returned.", relay_state)
+                plug_states.append(SmartStrip.SWITCH_STATE_UNKNOWN)
+
+        return plug_states
+
+    @state.setter
+    def state(self, value: str):
```
rytilahti, line 90: Let's make this to change the state of all the switches. Could you please add a new method `set_state(self, *, index=0)`, which should be used in-place of the state setter? We can keep the setter for adjusting all the plugs by simply calling that method with no parameters.

This will be the first step to add `set_*` instead of using property setters (so we can implement #98 and #119 properly).

pyHS100/smartstrip.py
```diff
@@ -0,0 +1,155 @@
...
+            raise ValueError("State %s is not valid.", value)
+
+    def is_on(self, index: int) -> bool:
+        """
+        Returns whether device is on.
+
+        :param index: plug index
+        :return: True if device is on, False otherwise
+        """
+        return self.plug[index].is_on()
```
rytilahti: This also should be `is_on(self, *, index: int = 0)` and returning a dictionary from all the plugs if no index is given.

rytilahti (commented):
Looking great! I added some comments, mainly:
1. All setters/getters should act on all sockets when no index is defined.
2. Fixing the situation with overcomplicated with proper set-methods (allows fixing #98 and #119 -- I think it's time to move away from property setters in the API in general to make the calls more explicit and easier to understand.)

newAM:
Sorry for the latency on my end, been busy at work.

I added `(self, *, index: int = -1)` indexing for functions.  I have the plugs zero indexed in the SmartStrip class for developers, and I do the one indexing translation in the CLI so that what the user sees the same plug index as printed on the device.

Any value less than zero will operate on all the plugs in the strip.

pyHS100/smartdevice.py
newAM: bad logic on my part there, I changed to to `if "child_num" in self.sys_info:`

pyHS100/smartstrip.py
```diff
@@ -41,34 +37,36 @@ def __init__(self,
...
+                switch_state = SmartPlug.SWITCH_STATE_OFF
             if relay_state == 1:
-                plug_states.append(SmartStrip.SWITCH_STATE_ON)
+                switch_state = SmartPlug.SWITCH_STATE_ON
             else:
                 _LOGGER.warning("Unknown state %s returned.", relay_state)
-                plug_states.append(SmartStrip.SWITCH_STATE_UNKNOWN)
+                switch_state = SmartPlug.SWITCH_STATE_UNKNOWN
+
+            plug_states.append({"index": plug, "state": switch_state})
```
rytilahti: This will create a list of dicts, I think having a dictionary (as indicated by the typing hint, too) would be nicer:
```
states = {}
for plug in range(self.num_children):
   ...
   states[plug] = relay_state
```
would be so much simpler, what do you think?

The same could be applied for the internal store of the plugs. Instead of storing a list of plugs in `self.plug`, having `self.plugs` dict with the index as a key would also simplify the iteration here?

pyHS100/smartstrip.py
```diff
@@ -79,49 +77,89 @@ def state(self, value: str):
...
+        :raises SmartDeviceException: on error
+        """
+        if not isinstance(value, str):
+            raise ValueError("State must be str, not of %s.", type(value))
+        elif value.upper() == SmartPlug.SWITCH_STATE_ON:
+            self.turn_on(index=index)
+        elif value.upper() == SmartPlug.SWITCH_STATE_OFF:
+            self.turn_off(index=index)
+        else:
+            raise ValueError("State %s is not valid.", value)
```
rytilahti: This should use `self.plugs`, like `self.plugs[index].set_state(value)`.

pyHS100/smartstrip.py
```diff
@@ -79,49 +77,89 @@ def state(self, value: str):
...
         """
-        return self.plug[index].is_on()
+        if index < 0:
+            is_on = []
+            for plug in range(self.num_children):
+                is_on.append({"index": plug,
+                              "is_on": self.plug[plug].is_on()})
+            return is_on
+        else:
+            return self.plug[index].is_on()
```
rytilahti: Same as above for internal dict store.

rytilahti (commented):
Please check out my proposal on storing the subplugs in a dict instead of a list.

newAM:
Sorry about that, I completely missed the mark there with the dict store.  It's painfully clear what you were after originally.  If everything is alright now I'll start writing the tests.

pyHS100/smartstrip.py
```diff
@@ -0,0 +1,190 @@
...
+
+        :returns: list with the state of each child plug
+                  SWITCH_STATE_ON
+                  SWITCH_STATE_OFF
+                  SWITCH_STATE_UNKNOWN
+        :rtype: dict
+        """
+        states = {}
+        for plug in range(self.num_children):
+            relay_state = self.sys_info["children"][plug]["state"]
```
rytilahti: I would still prefer if `self.plugs` were a dict instead of a list (makes also `if X in` comparisons nicer when checking if a plug exists, see my commentary below), making `for plug in self.plugs` is preferable here. Please move calling `self.sysinfo` before the loop to avoid unnecessary I/O (Generally speaking I think we should cache `sysinfo` and refresh the cache where necessary, but that's for another PR..).

pyHS100/smartstrip.py
```diff
@@ -0,0 +1,190 @@
...
+        elif value.upper() == SmartPlug.SWITCH_STATE_ON:
+            self.turn_on()
+        elif value.upper() == SmartPlug.SWITCH_STATE_OFF:
+            self.turn_off()
+        else:
+            raise ValueError("State %s is not valid.", value)
+
+    def set_state(self, value: str, *, index: int = -1):
+        """
+        sets the state of a plug on the strip
```
rytilahti:
```suggestion
        Sets the state of a plug on the strip
```

pyHS100/smartstrip.py
```diff
@@ -0,0 +1,190 @@
...
+
+        :param value: one of
+                    SWITCH_STATE_ON
+                    SWITCH_STATE_OFF
+        :param index: plug index (-1 for all)
+        :raises ValueError: on invalid state
+        :raises SmartDeviceException: on error
+        """
+        self.plugs[index].set_state(value)
```
rytilahti: Bounds check is missing. Please add `SmartStripException(SmartDeviceException)` and raise it here if the index is not found in the `self.plugs` (`if index in self.plugs` will work when `self.plugs` is an index-keyed dict, please add tests for out-of-bound exceptions, too :-)).

pyHS100/smartstrip.py
```diff
@@ -0,0 +1,190 @@
...
+        :param index: plug index (-1 for all)
+        :return: True if device is on, False otherwise, Dict without index
+        """
+        if index < 0:
+            is_on = {}
+            for plug in range(self.num_children):
+                is_on[plug] = self.plug[plug].is_on()
+            return is_on
+        else:
+            return self.plugs[index].is_on()
```
rytilahti: Same as above for index checks.

pyHS100/smartstrip.py
```diff
@@ -0,0 +1,190 @@
...
+        """
+        Returns whether device is on.
+
+        :param index: plug index (-1 for all)
+        :return: True if device is on, False otherwise, Dict without index
+        """
+        if index < 0:
+            is_on = {}
+            for plug in range(self.num_children):
+                is_on[plug] = self.plug[plug].is_on()
```
rytilahti: Instead of calling this once per plug, this information should be read once from the sysinfo to avoid unnecessary I/O.

pyHS100/smartstrip.py
```diff
@@ -0,0 +1,190 @@
...
+        """
+        Turns outlets on
+
+        :param index: plug index (-1 for all)
+        :raises SmartDeviceException: on error
+        """
+        if index < 0:
+            self._query_helper("system", "set_relay_state", {"state": 1})
+        else:
+            self.plugs[index].turn_on()
```
rytilahti, line 159: Bound checks.

pyHS100/smartstrip.py
```diff
@@ -0,0 +1,190 @@
...
+        Returns pretty-printed on-time
+
+        :param index: plug index (-1 for all)
+        :return: datetime for on since
+        :rtype: datetime with index, Dict[int, str] without index
+        """
+        if index < 0:
+            on_since = {}
+            for plug in range(self.num_children):
+                on_since[plug] = self.plugs[plug].on_since
```
rytilahti: See my comment about fetching the info only once.

pyHS100/smartstrip.py
```diff
@@ -0,0 +1,190 @@
...
+        """
+        Returns strip-specific state information.
+
+        :return: Strip information dict, keys in user-presentable form.
+        :rtype: dict
+        """
+        state = {'LED state': self.led}
+        for plug_index in range(self.num_children):
+            state['Plug %d on since' % (plug_index + 1)] = \
+                self.on_since(index=plug_index)
```
rytilahti: This can also be fetched once.

rytilahti (commented):
It is looking good! I added a couple of comments, mainly related to error-handling and unnecessary I/O. For tests it is only necessary to test the smartstrip-specific functionality, no need to rewrite the SmartPlug related tests again.

newAM:
I added the tests, they are still a work in progress since they will not pass with the `FakeTransportProtocol`, but they do pass on real hardware.

Some interesting notes:
- Latitude and Longitude changed for HS300, they are no longer floats.
- `for x in range(self.num_children):` is a noticeable amount faster than `for x in self.plugs:`.  I do not have any conclusive numbers yet since test times on the real hardware varies quite a bit.
- icons are elusive, they don't seem to be stored under the children or the strip itself.  I'll have to setup a capture to find out if they are still stored on the device.

jimboca:
So do we need to finish the tests before this can be merged?

kirichkov:
> So do we need to finish the tests before this can be merged?

Sadly - yes. Otherwise when someone else makes changes in other parts of the code they can unintentionally break the HS300.

newAM:
The test transport protocol should not be too difficult to update, I plan to finish it this weekend (barring any work emergency which there have been a lot of recently 😩)

requirements.txt
```diff
@@ -1,2 +1,2 @@
 click
-click-datetime
\ No newline at end of file
+click-datetime
```
rytilahti, line 2: I'll do a review later this week, but just saw this commit in my e-mail and thought to add that there should be no need for click-datetime at all anymore, as such functionality is offered directly by click: https://click.palletsprojects.com/en/7.x/parameters/#parameter-types (search for datetime).

jimboca: ok, I'm not familiar with it, I was just testing and it crashed due to that missing module so I added it to the list.

rytilahti: Ok, that's fine, it can be cleaned out later as deemed necessary. I did a cursory read-through and I think it's best if we merge this now. I've been working to simplify the API and tests, so I hope I'll get that done sooner than later and we will then prepare a new release.

Thanks a lot to you and @newAM for working on this, it's much appreciated and awaited feature! :-)

jimboca:
Awesome!  Thanks to you and especially @newAM who did most, pretty much all, of the work.  Can't wait to integrate this into an ISY Nodeserver.

newAM:
and thank you guys for doing the review!  I am excited to get this into home assistant! 👍

theclevercrib:
Hi guys!  Is there any chance we'll find out when the next release will be?  I would really love to get this working inside of my Home Assistant configuration! Thanks.

PR #140, opened 2018-11-16
Change valid hue range to 0-359 (fixes #130), update README.md & test

### LINKED ISSUE #130: bulb hue is limited to 255 instead of 360
Unless this was intentional because you don't want people having pink lights ¯\_(ツ)_/¯
I tried to make a pr but I don't have access. Here's a diff for the fix
```
diff --git a/pyHS100/smartbulb.py b/pyHS100/smartbulb.py
index 8f0bce6..5591974 100644
--- a/pyHS100/smartbulb.py
+++ b/pyHS100/smartbulb.py
@@ -142,10 +142,10 @@ class SmartBulb(SmartDevice):
         if not self.is_color:
             return None
 
-        if not isinstance(state[0], int) or not (0 <= state[0] <= 255):
+        if not isinstance(state[0], int) or not (0 <= state[0] <= 360):
             raise SmartDeviceException(
                     'Invalid hue value: {} '
-                    '(valid range: 0-255)'.format(state[0]))
+                    '(valid range: 0-360)'.format(state[0]))
 
         if not isinstance(state[1], int) or not (0 <= state[1] <= 100):
             raise SmartDeviceException(
```

annikajayuk:
Tested on LB130(EU) hardware 1.0, firmware 1.8.6 Build 180809 Rel.091659

Co-Authored-By: @nkonopinski <nkonopinski@users.noreply.github.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,7 +206,7 @@ if bulb.is_variable_color_temp:
 
 ### Setting the color
 
-Hue is given in degrees (0-360) and saturation and value in percentage.
+Hue is given in degrees (0-359) and saturation and value in percentage.
 
 ```python
 print(bulb.hsv)
--- a/pyHS100/smartbulb.py
+++ b/pyHS100/smartbulb.py
@@ -142,10 +142,10 @@ class SmartBulb(SmartDevice):
         if not self.is_color:
             return None
 
-        if not isinstance(state[0], int) or not (0 <= state[0] <= 255):
+        if not isinstance(state[0], int) or not (0 <= state[0] <= 359):
             raise SmartDeviceException(
                     'Invalid hue value: {} '
-                    '(valid range: 0-255)'.format(state[0]))
+                    '(valid range: 0-359)'.format(state[0]))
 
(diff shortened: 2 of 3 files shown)


rytilahti:
Thanks, looks good! Could you also provide an example where the hue is above 255 (to be included in fakes.py)?

annikajayuk:
> Thanks, looks good! Could you also provide an example where the hue is above 255 (to be included in fakes.py)?

I'm sorry, I'm not sure what you mean; an example of what? A colour where the hue is above 255? Pink would be 300 degrees hue, 100% saturation, 100% brightness, but I can't see where this would go in fakes.py (sorry, I'm just hacking about here, not experienced).

Thanks,
Annika

rytilahti:
Actually, I'll take that back. I was just wondering that it'd be a great idea to have real outputs (including all the details) from devices for testing, but that can wait if/when the tests get cleaned up (instead of having those in variables inside fakes.py they should be simply json files containing the output). Let's get this merged now, I'll prepare a release when #131 gets merged :-)

PR #142, opened 2018-11-23
cli: fix hsv get

annikajayuk:
Raising an exception if an incomplete HSV colour is provided.

### DIFF
--- a/pyHS100/cli.py
+++ b/pyHS100/cli.py
@@ -219,7 +219,7 @@ def hsv(dev, ctx, h, s, v):
     if h is None:
         click.echo("Current HSV: %s %s %s" % dev.hsv)
     elif s is None or v is None:
-        raise click.BadArgumentUsage("Incomplete HSV color provided.", ctx)
+        raise click.BadArgumentUsage("Changing color needs 3 values.", ctx)
     else:
         click.echo("Setting HSV: %s %s %s" % (h, s, v))
         dev.hsv = (h, s, v)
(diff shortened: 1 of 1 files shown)


pyHS100/cli.py
```diff
@@ -209,14 +209,17 @@ def temperature(dev, temperature):
...
 @pass_dev
-def hsv(dev, h, s, v):
+def hsv(dev, ctx, h, s, v):
     """Get or set color in HSV. (Bulb only)"""
-    if h is None or s is None or v is None:
-        click.echo("Current HSV: %s" % dev.hsv)
+    if h is None:
+        click.echo("Current HSV: %s %s %s" % dev.hsv)
+    elif s is None or v is None:
+        raise click.BadArgumentUsage("Setting a color requires 3 values.", ctx)
```
rytilahti, line 222: What happens here if no context is passed? Otherwise this looks to be merged, thanks!

annikajayuk: Without passing the context, the line
`Try "pyhs100 hsv --help" for help.`
isn’t printed, which _is_ included when Click raises an exception automatically (for example when a value is outside the specified range). I don’t know why, because the first line is
`Usage: pyhs100 hsv [OPTIONS] [H] [S] [V]`
even without the context being passed, so Click obviously knows where the exception was raised in either case.

rytilahti: Ok, let's merge it, thanks again :-)

PR #150, opened 2019-01-03
avoid 'referenced before assignment' exception

tripzero:
On occasion I see connectivity issues which triggers exceptions on socket.create_connection failing and then trying to call close() on sock.


rytilahti:
LGTM, thanks!

PR #156, opened 2019-01-26
Add kelvin range for KL130

dieselrabbit:
This addition gets this bulb to work with home assistant.


kirichkov (approved):
🎉

rytilahti:
Could you also add the range for KL120 (seems to be the same as LB120: https://www.tp-link.com/uk/products/details/cat-5609_KL120.html) to this same commit?

dieselrabbit:
Added KL120, though I can't test as I don't have that bulb.
Hope I did this right. Git newbie and this is my first pull request.

rytilahti:
It is good, thanks! :1st_place_medal:

PR #159, opened 2019-01-30
Fix Discover#discover incorrect documentation

kirichkov:
The documentation states the timeout defaults to 5 seconds, but in the definition of the method timeout is set to 3


rytilahti:
Good catch! Weird that the coverage increased by this commit though.. Let's merge it! :-)

kirichkov:
> Good catch! Weird that the coverage increased by this commit though.. Let's merge it! :-)

That must be some special feature when the commit-s message is longer than the changes themselves 😄

PR #161, opened 2019-02-26
Smartstrip: return on_since state information only when the socket is on

rytilahti:
`on_since` information is added to the state attributes only when the socket is on.
This PR also includes adding proper cli printouts for the child sockets.

Fixes #160

PR #151, opened 2019-01-09
API and tests cleanup

rytilahti:
I'm creating this PR to get some feedback on the proposed API changes. The current API remains intact (with some minor changes related to error-reporting and available constants), but I'm hoping we can drop the property-setter based API in the future. The main goal was to make the API consistent and nicer to use for 3rd parties, while making writing tests more manageable. This also allows easier implementation of some previously wished features.

This is still WIP (HS300 has to be rebased in, and tests need to be made runnable on real devices) before merging it. The version should be bumped to 0.4.0 with these changes.

Feel free to test and contribute new fixture files (`pyhs100 dump-discover` can be used to do that).

Related issues #97, #146,  #147 and PRs #145, #119, #98.

The commit messages explaining the changes can be found below.

    Cleanup tests and improve test coverage
    
    * Make writing tests easier by sharing code for common implementations
    * Instead of storing test data inside python files, dump-discover based information is used
      * This will simplify adding new tests and remove code duplication
      * fixtures are based on https://github.com/plasticrake/tplink-smarthome-simulator

    Major API cleanup
    
    Properties shall no more change the state of the device, this work in still in progress, the main goal being making the API more user-friendly and to make implementing new features simpler.
    
    The newly deprecated functionality will remain working and will simply warn the user about deprecation.
    Previously deprecated 'features' property and 'identify' method are now finally removed.
    
    Deprecate and replace the following property setters:
    * state with turn_on() and turn_off()
    * hsv with set_hsv()
    * color_temp with set_color_temp()
    * brightness with set_brightness()
    * led with set_led()
    * alias with set_alias()
    * mac with set_mac()
    
    And getters:
    * state with is_on and is_off
    
    The {BULB,PLUG}_STATE_{ON,OFF} is simplified to STATE_ON and STATE_OFF, UNKNOWN state is removed.
    These are now deprecated and will be removed in the future.
    * is_on and is_off can be used to check for the state
    * turn_on() and turn_off() for changing the device state.
    
    Trying to use functionality not supported by the device will cause SmartDeviceExceptions instead of failing silently and/or returning None.
    This includes, e.g., trying to set a color temperature on non-supported bulb.
    
    ValueErrors are raised instead of SmartDeviceExceptions where appropriate (e.g. when trying to set an invalid hsv or brightness).
    
    New enum type DeviceType is added to allow detecting device types without resorting to isinstance() calling. SmartDevice class' device_type property can be used to query the type. is_plug and is_bulb helpers are added.

[...]

PR #162, opened 2019-02-27
Enhancement: Added parameter to allow discovery of certain device types.

zemerick1:
Device types are based on the classes that are returned by _get_device_class().

Example: print(Discover().discover(deviceType="SmartPlug"))

### DIFF
--- a/pyHS100/discover.py
+++ b/pyHS100/discover.py
@@ -62,7 +62,11 @@ class Discover:
                     devices[ip] = device_class(ip)
                 else:
                     return None
-
+        except socket.timeout:
+            _LOGGER.debug("Got socket timeout, which is okay.")
+        except Exception as ex:
+            _LOGGER.error("Got exception %s", ex, exc_info=True)
+        return devices
     @staticmethod
     def discover_single(host: str,
                         protocol: TPLinkSmartHomeProtocol = None
--- a/pyHS100/discover.py
+++ b/pyHS100/discover.py
@@ -16,7 +16,7 @@ class Discover:
     @staticmethod
     def discover(protocol: TPLinkSmartHomeProtocol = None,
                  port: int = 9999,
-                 timeout: int = 3, 
+                 timeout: int = 3,
                  deviceType: str = None) -> Dict[str, SmartDevice]:
         """
         Sends discovery message to 255.255.255.255:9999 in order
@@ -60,13 +60,12 @@ class Discover:
                             devices[ip] = device_class(ip)
                 elif device_class is not None:
                     devices[ip] = device_class(ip)
-                else:
-                    return None
         except socket.timeout:
             _LOGGER.debug("Got socket timeout, which is okay.")
         except Exception as ex:
             _LOGGER.error("Got exception %s", ex, exc_info=True)
         return devices
+
     @staticmethod
     def discover_single(host: str,
                         protocol: TPLinkSmartHomeProtocol = None
--- a/pyHS100/discover.py
+++ b/pyHS100/discover.py
@@ -16,7 +16,8 @@ class Discover:
     @staticmethod
     def discover(protocol: TPLinkSmartHomeProtocol = None,
                  port: int = 9999,
-                 timeout: int = 3) -> Dict[str, SmartDevice]:
+                 timeout: int = 3, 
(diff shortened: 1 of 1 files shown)


pyHS100/discover.py
```diff
@@ -53,14 +54,19 @@ def discover(protocol: TPLinkSmartHomeProtocol = None,
...
                 device_class = Discover._get_device_class(info)
-                if device_class is not None:
+                if deviceType is not None:
+                    if deviceType in str(device_class):
+                        if device_class is not None:
+                            devices[ip] = device_class(ip)
+                elif device_class is not None:
                     devices[ip] = device_class(ip)
+                else:
+                    return None
```
kirichkov: This line looks like a breaking change. Before we did not return `None` but the `devices` variable in https://github.com/GadgetReactor/pyHS100/blob/07b4eda7ae3250c3df826de6f1dbd6b6c5f873ef/pyHS100/discover.py#L62

In turn devices is initialized as an empty dict https://github.com/GadgetReactor/pyHS100/blob/07b4eda7ae3250c3df826de6f1dbd6b6c5f873ef/pyHS100/discover.py#L47

pyHS100/discover.py
```diff
@@ -53,14 +54,19 @@ def discover(protocol: TPLinkSmartHomeProtocol = None,
...
         except socket.timeout:
             _LOGGER.debug("Got socket timeout, which is okay.")
         except Exception as ex:
             _LOGGER.error("Got exception %s", ex, exc_info=True)
         return devices
-
```
kirichkov: Please revert this empty line deletion. It's good to have it for readability.

kirichkov (changes requested):

pyHS100/discover.py
zemerick1: It's not really a breaking change since it only returns None if you specify the deviceType. The original return is still there. However, I will change it to be consistent.

zemerick1:
Requested changes added.

zemerick1:
Bump.
Is anyone still maintaining this repo? I see 1 year old PRs

rytilahti:
Yes, this is still maintained albeit slowly (considering it is pretty mature already). Considering this is just change to the discovery API, you may want to wait until #151 lands into the repository as it will add `DeviceType` enum which would allow nicer API for this.

pyHS100/discover.py
```diff
@@ -16,7 +16,8 @@ class Discover:
     @staticmethod
     def discover(protocol: TPLinkSmartHomeProtocol = None,
                  port: int = 9999,
-                 timeout: int = 3) -> Dict[str, SmartDevice]:
+                 timeout: int = 3,
+                 deviceType: str = None) -> Dict[str, SmartDevice]:
```
rytilahti, line 20: `device_type` is the pythonic form.

pyHS100/discover.py
```diff
@@ -53,7 +54,11 @@ def discover(protocol: TPLinkSmartHomeProtocol = None,
                 ip, port = addr
                 info = json.loads(protocol.decrypt(data))
                 device_class = Discover._get_device_class(info)
-                if device_class is not None:
+                if deviceType is not None:
```
rytilahti, line 57:
```suggestion
                if device_type is not None and device_class is not None:
```
to avoid another layer of nesting below. See my comment in the PR though, I think it will be much better when the type is defined as an enum.

zemerick1:
Do we know when that is due to be released?

pyHS100/discover.py
zemerick1: bad habits die hard. I'll just wait on the adjustments mentioned in the comments.

rytilahti:
No idea, I think the only point I'm not satisfied with is how the smartstrip sockets are exposed (I think it has to be moved away from overloading the `set_*`, my bad for pushing for that). But other than that, the code itself should be in a testable state and I would like to hear feedback on potential issues.

edit: before getting it merged though I plan to make a single release before the API change, which currently waits for #163 and #161 to be merged.

rytilahti:
Would you like to update this PR on top of the current master? There is now a way to get the device type enum & query the type with `is_` methods, too.

rytilahti:

Issue #165, opened 2019-03-14
Crash when __repr__ is called on SmartStrip device


scootdj:
SmartDevice defines a __repr__ that calls self.is_on.  

is_on must be implemented by subclasses.  SmartStrip defines is_on differently with an additional parameter (index) and a different return value 'Any'. 

Infinite recursion happens when __repr__ tries to call the SmartStrip.is_on.  Because the value returned by SmartDevice.is_on is not a string, a recursive call to __repr__ happens.

I have worked around it by changing SmartDevice.__repr__ to:

def __repr__(self):
        if isinstance(self.is_on, str):
            is_on = self.is_on
        else:
            is_on = self.is_on()
        return "<%s at %s (%s), is_on: %s - dev specific: %s>" % (
            self.__class__.__name__,
            self.host,
            self.alias,
            is_on,
            self.state_information)

jimboca:
This issue looks like it's fixed in @rytilahti api_and_tests_cleanup branch.  Can we get that merged in and released?

eode:
Cleaned that up for copypasta, if anyone wants it.
```python
def fixed_repr(self):
    if isinstance(self.is_on, str):
        is_on = self.is_on
    else:
        is_on = self.is_on()
    return "<%s at %s (%s), is_on: %s - dev specific: %s>" % (
        self.__class__.__name__,
        self.host,
        self.alias,
        is_on,
        self.state_information)

pyHS100.SmartStrip.__repr__ = fixed_repr
```

newAM:
I know #151 fixes this, but I am not sure what the timeline is like on that so I created #169 as a targeted fix for this, plus unit tests to catch this in the future.

PR #169, opened 2019-04-06
Fix for SmartStrip repr

newAM:
Targeted fix for issue #165, with unit tests to catch this in the future.


rytilahti:
Looks good to me, thanks for the PR! I'll try to make a final release before the API breaking #151, hopefully already this week :+1:

PR #174, opened 2019-06-02
Fix temperature range for KL120, the device can only do up to 5000 k

rytilahti:
I'm not definitely sure about this, as https://www.bhphotovideo.com/c/product/1433489-REG/tp_link_kl120_kasa_smart_light.html lists 5000K as the maximum (and my own experiences confirm that), but https://www.tp-link.com/pt/home-networking/smart-bulb/kl120/ lists that it'd support up to 6500k.

I have the ones from that b&h link, and they will report an error when trying to set anything above 5000 (see #173 for a related PR). It reports itself as KL120(US) with the following versions:
```
Hardware:     1.0
Software:     1.8.6 Build 180809 Rel.091659


rytilahti:
Looks like this is a difference between US & EU bulbs, see https://www.kasasmart.com/us/products/smart-lighting/kasa-smart-wi-fi-light-bulb-tunable-white-kl120 for the US version (which correctly talks about 5000k). I'll update this PR accordingly.

rytilahti:
I suppose this is fine, merging.

Issue #158, opened 2019-01-30
Discover not working after period of time


ElliotReeve:
I am using a raspberry pi to read temperature of the room once an hour. If it’s too cold I turn on the plug which has a heater attached.

The approach I took was to discover the devices first to grab the IP address dynamically then use that to connect to the plug.

However, I have noticed that the discover functionality fails to work after say 6 or so hours.

Any ideas?

kirichkov:
Have you checked whether your plug is working at all? I had problems with my HS110 simply crashing when the line voltaged dropped. I never figurred out how significant the drop should be for it to crash, but I assume something like 180V (from 220-240V) just make the plug to crash. It's not accessible via the TP-Link either, doesn't respond to pings, etc. Only plugging it out of the socked and plugging it back in fixed it - i.e. restarting it.

I'd suggest whenever this happens next time you try pinging it and report back.

ElliotReeve:
Yeah so if I run discover I get nothing back.

But if i manually put in the IP that I no the plug is connected to I get the full info.

```
for dev in Discover.discover().values():
    print(dev)
```

Returns nothing

```
plug = SmartPlug("192.168.1.4")
print(plug.get_sysinfo())
```

Returns

`{'sw_ver': '1.5.8 Build 180815 Rel.135935', 'hw_ver': '2.0', 'type': 'IOT.SMARTPLUGSWITCH', 'model': 'HS100(UK)', 'mac': 'AC:84:.......
`

kirichkov:
Can you use `discover` but increasing the timeout and setting it to 10, 20 or even 60? Perhaps your plug is not responding in the 3 seconds allocated for timeout?

You can try sniffing with WireShark whether the plug is responding to the discovery requests - they should be sent from your raspberry to IP 255.255.255.255 port 9999. tcpdump/ngrep will also work.

sylvandb:
Also consider, does your raspberry pi have two network connections (e.g. ethernet and WiFi)?

Discovery with two network connections needs to use the appropriate network broadcast address, not the global broadcast. The global broadcast is not network specific and will only go out one network, maybe not the correct one.

For example, if you have 192.168.0.X/24 (netmask 255.255.255.0) then a broadcast of 192.168.0.255 will discover on that network.  If you also have 172.30.123.45/24 then broadcast 172.30.123.255 to discover.

avanshee:
[...]

PR #176, opened 2019-07-20
Specify broadcast address for discovery

kochinc:
Sometimes it's useful to be able to specify the broadcast address for discovery. On a host with multiple network interfaces, when a discovery message is sent to 255.255.255.255:9999, the routing table decides which interface it is routed to. By specifying broadcast address, e.g. 192.168.xxx.255, the discovery message can be directed to a particular subnet only.

The modification to discover.py is very limited and shouldn't affect existing programs using the library. The cli.py is changed to include a --target option for specifying broadcast address.

### DIFF
--- a/pyHS100/discover.py
+++ b/pyHS100/discover.py
@@ -65,8 +65,6 @@ class Discover:
         if protocol is None:
             protocol = TPLinkSmartHomeProtocol()
 
-        #target = "255.255.255.255"
-
         sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
         sock.setsockopt(socket.SOL_SOCKET, socket.SO_BROADCAST, 1)
         sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
--- a/pyHS100/discover.py
+++ b/pyHS100/discover.py
@@ -43,6 +43,7 @@ class Discover:
     @staticmethod
     def discover(
         protocol: TPLinkSmartHomeProtocol = None,
+        target: str = "255.255.255.255",
         port: int = 9999,
         timeout: int = 3,
         discovery_packets=3,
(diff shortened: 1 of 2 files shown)


pyHS100/discover.py
```diff
@@ -63,7 +65,7 @@ def discover(
         if protocol is None:
             protocol = TPLinkSmartHomeProtocol()
 
-        target = "255.255.255.255"
+        #target = "255.255.255.255"
```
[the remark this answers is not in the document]
kochinc: That line can be deleted.

kirichkov (approved):
Looks good to me.

@rytilahti merge if you're OK.

pyHS100/discover.py
```diff
@@ -43,6 +43,7 @@ class Discover:
     @staticmethod
     def discover(
         protocol: TPLinkSmartHomeProtocol = None,
+        target: str = "255.255.255.255",
```
rytilahti, line 46: I'm not sure if that's the proper position for this in the argument list, but considering the current master has evolved and is breaking the backwards compatibility, I'd propose that we simply adjust discover to allow keyword-only arguments.

rytilahti (commented):
Looks fine to me, I have only one question: do you think it's better to have an address instead of an interface name for this?

pyHS100/discover.py
rytilahti: After giving this a bit thought, I think this can be changed in a separate PR, so let's get this merged if someone is eagerly waiting for this functionality.  #158 is at least partially related (there are mentions about multiple subnets in the comments).

kirichkov:
> Looks fine to me, I have only one question: do you think it's better to have an address instead of an interface name for this?

Not sure whether this was directed at me, or @kochinc, my general preference is to deal with interfaces, rather than IP addresses - network ranges change, interfaces generally stay the same. If go for interfaces, however, how do we address the 'broadcast' since broadcast should send to all interfaces? Perhaps a `interface='all'` keyword argument?

kochinc:
Either way it's fine for me. At first I wanted to add an interface parameter. After inspected the code, I realized it's very simple to just include a broadcast target address parameter. That solved my issue with TP-Link device discovery on Home Assistant.

jkriegshauser:
Would it be possible to get a release with this merged so that we can get Home Assistant working with Kasa on different subnets before python-kasa is integrated?

PR #178, opened 2019-07-26
Fix __repr__ function of SmartDevice

bmoregeo:
The following fails due to __repr__ function on the device is missing… a format parameter. This adds in a test and fixes the bug

for dev in Discover.discover().values():
    print(dev)


rytilahti (approved):
LGTM, thanks!

PR #180, opened 2019-08-01
Python < 3.5 cleanups

scop:
Python 3.5 is said to be the minimum version, so clean up things that are there for earlier versions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,7 +8,7 @@ setup(name='pyHS100',
       author_email='sean@gadgetreactor.com',
       license='GPLv3',
       packages=['pyHS100'],
-      install_requires=['click', 'typing', 'deprecation'],
+      install_requires=['click', 'deprecation'],
       python_requires='>=3.5',
       entry_points={
             'console_scripts': [
--- a/tox.ini
+++ b/tox.ini
@@ -1,10 +1,8 @@
 [tox]
-envlist=py33,py34,py35,py36,py37,flake8
+envlist=py35,py36,py37,flake8
 skip_missing_interpreters = True
 
 [tox:travis]
-3.3 = py33
-3.4 = py34
 3.5 = py35
 3.6 = py36
 3.7 = py37


rytilahti:
Thanks for the PR! I've actually been thinking if it'd be time to move on to depend on even newer python versions, the newest Debian stable being out with python 3.7 after all, and the most profilic user of this library, homeassistant, is also moving away from supporting 3.5. I think f-strings (introduced in 3.6) would be at least nice to have. Any opinions on bumping the version requirement for the next major release (which will already introduce API changes)?

scop:
None at all here, but then again I'm not using this library at all at the moment. The reason that triggered this PR was that home assistant's CI setup was somehow negatively affected by typing getting pulled in by a couple of dependencies. There is a workaround in HA in place now, but I filed PR's to get rid of the deps in upstream projects so it'll be fixed cleaner later.

rytilahti:
Great, let's get this merged anyway :+1:

matlimatli:
Would it be possible to get this fix included in a PyPI release? It seems like the dependency on typing may still cause some issues with the testing environment for home-assistant.

rytilahti:
The master branch of this project has already been diverged so much from the last release, that I don't feel confident in touching any of it anymore. Also, this lib is going to be replaced by python-kasa (https://github.com/python-kasa/python-kasa) in the near future for homeassistant, so I have personally no interest going through the hassle to prepare a new release.

matlimatli:
I see. @scop Do you have any more information about the workaround mentioned above?

scop:
On Sat, 23 May 2020 at 23:30, matlimatli <notifications@github.com> wrote:

> I see. @scop <https://github.com/scop> Do you have any more information
> about the workaround mentioned above?
>
Not readily at hand. But if you search for Home Assistant issues and PR's
from around the time this was filed, it's possible some traces are found
there. Unfortunately in HA commit messages do not tend to contain much
details or useful info, so I doubt there's anything there. Perhaps balloob
or pvizeli remember something.

I also believe the issue is not a transient one -- ISTR the related
workaround was recently added to some ~new CI jobs in addition to still
being around in the old ones.

PR #182, opened 2019-10-25
Fix cli status reporting, blackify cli.py

### LINKED ISSUE #181: bug displaying plug status if off.
when running pyhs100 if the device is on you get :-  == Piaware - HS110(UK) ==
Device state: ON
Host/IP: 192.168.1.67

however if the device is off you get :- 
== Empty 1 - HS110(UK) ==
OFF
Host/IP: 192.168.1.188

seems in cli.py the line :- 
"Device state: %s" % "ON" if dev.is_on else "OFF",

needs changing to :- 

"Device state: %s" % ("ON" if dev.is_on else "OFF"),

rytilahti:
Fixes #181

### DIFF
--- a/pyHS100/cli.py
+++ b/pyHS100/cli.py
@@ -42,7 +42,8 @@ pass_dev = click.make_pass_decorator(SmartDevice)
 )
 @click.option(
     "--target",
-    default="255.255.255.255", required=False,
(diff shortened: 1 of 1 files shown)


kirichkov (approved):
Good to go!

PR #184, opened 2019-11-11
use asyncio

basnijholt:
What about implementing this library using `asyncio`? I would like to know what you guys think? In this way, device communication will be MUCH faster! 

It breaks backward compatibility because many `property`s that are implemented actually should never have been properties (because there is a computation involved) and have to be awaited.

Because of the many breaking changes, I have removed the marked `deprecated` methods too.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,0 @@
-[flake8]
-ignore = E203, E266, E501, W503, F403, F401
-max-line-length = 79
-max-complexity = 18
-select = B,C,E,F,W,T4,B9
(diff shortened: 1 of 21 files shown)


basnijholt:
I now implemented all the devices and all but the `test_cache` and `test_cache_invalidates` are passing.

It is not that the cache is actually broken, but that the test just doesn't work.

Does anyone understand why that happens?

basnijholt:
@rytilahti, I merged your #185 branch! 👍

basnijholt:
Closing in favour of the new repository and organisation: https://github.com/python-kasa/python-kasa

PR #188, opened 2019-12-20
Updates and bugfixes for KL130

ervanalb:
This pull request adds several features:

- Commands to get and set wifi network (useful for initial provisioning without app)
- Commands to get cloud info and disable cloud connectivity (for privacy)
- Command to set location

It also fixes the following problems with my KL130:

- Set alias does not work
- Get location does not work
- Get time does not work

I put these changes into the `SmartBulb` class, assuming all bulbs are like my KL130. If that is not the case, it may break these commands. I don't have a way to test other devices. The SmartPlugs should be unaffected by these changes (other than adding new commands.)


kirichkov:
Hi There! Thanks for your contribution. We're currently doing some reorganization and moving the code to a new organization. Would it be possible for you to check your code against this repo:

https://github.com/python-kasa/python-kasa

PR #197, opened 2020-11-24
Removes emeter discovery

bmbouter:
This caused interoperatbility isssues with newer HS220 hardware.
Removing this has fixed it for several folks.

See https://github.com/home-assistant/core/issues/39395

### DIFF
--- a/pyHS100/discover.py
+++ b/pyHS100/discover.py
@@ -10,8 +10,7 @@ _LOGGER = logging.getLogger(__name__)
 
 
 class Discover:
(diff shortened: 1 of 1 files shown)


rytilahti:
Considering that the emeter response information was never used in any released version, I think it's fair enough merge this and do a 0.3.5.2 release.

I'll update the README to point out that the `goodbye_release` (and not master) contains the newest pypi published code, as I already got confused earlier when trying to figure out if it'd be okay to merge this or not (checking the code in my local master checkout).

PR #201, opened 2021-02-23
Add KL125 definition

mdarnol:
Recently purchased some KL125s and noticed that they caused home assistant to crash because they were not defined and there was no default color temp

Product Link
https://www.kasasmart.com/us/products/smart-lighting/kasa-smart-light-bulb-multicolor-kl125

It says there that the color temp is 2500K-6500K (adjustable via app and cloud)

Reason for the PR
```
2021-02-23 00:21:44 ERROR (MainThread) [homeassistant.components.light] tplink: Error on device update!
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/entity_platform.py", line 358, in _async_add_entity
    await entity.async_device_update(warning=False)
  File "/usr/src/homeassistant/homeassistant/helpers/entity.py", line 466, in async_device_update
    await task
  File "/usr/src/homeassistant/homeassistant/components/tplink/light.py", line 468, in async_update
    is_ready = await self.hass.async_add_executor_job(
  File "/usr/local/lib/python3.8/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/src/homeassistant/homeassistant/components/tplink/light.py", line 248, in attempt_update
    self._light_features = self._get_light_features()
  File "/usr/src/homeassistant/homeassistant/components/tplink/light.py", line 289, in _get_light_features
    min_mireds = kelvin_to_mired(min_range)
  File "/usr/src/homeassistant/homeassistant/util/color.py", line 515, in color_temperature_kelvin_to_mired
    return math.floor(1000000 / kelvin_temperature)
ZeroDivisionError: division by zero
```

PR #202, opened 2021-03-22
Add support for KL125

superm1:
(Yes; I realize that this repo is abandoned and stuff should go into python-kasa now)

I recently purchased KL125's and like many others found they didn't work with Home Assistant due to a divide by zero error.  There was previously this submitted https://github.com/GadgetReactor/pyHS100/pull/201 to fix it but that request was closed and indicated to go over to python-kasa.

I saw this, and I checked with `python-kasa` and things _work_ for my KL125's.  However there isn't yet a component in home assistant to use `python-kasa` instead.  So I know that's the eventual direction, but considering this, i'm wondering if you would consider to merge and tag another release with this very minor change.  If so I'll work out getting Home Assistant bumped up to the new tag.


WesForster:
I'm in the same boat having bought the KL125s, as well.

superm1:
> I'm in the same boat having bought the KL125s, as well.  

FWIW it's possible to run the forked dependency and everything "works".

rytilahti:
There's a PR to fix alleviate this problem on homeassistant: https://github.com/home-assistant/core/pull/48235 so I'd rather avoid doing any new releases for this project.

